PR #48, opened 2025-08-11
Add some xngin-specific names to the Sentry denylists.

dctaf:
This removes some Evidential-specific names from Sentry error reports.


qixotic (approved):
Thanks! 
BTW, our tq cli also uses sentry. Wouldn't hurt to apply the same settings to it as well for consistency right (although maybe database_url at best might show up from there)?

PR #51, opened 2025-08-14
fix assignment bug when the participant_id column uses big integers

qixotic:
## Description

Stochatreat’s internals adds fake rows of data as a step, which introduces nulls to the unique id column. If it is of certain dtypes, pandas will upcast it to a float, losing precision and cause us to mismatch/drop treatment assignments.

Fix is a workaround that pre-casts the column type to object directly, avoiding the loss of precision.

## How has this been tested?

Unittests added to first verify failures before fix, then passing after.

## Checklist

Fill with `x` for completed.

- [* ] I have updated the automated tests

### DIFF
--- a/src/xngin/stats/assignment.py
+++ b/src/xngin/stats/assignment.py
@@ -10,6 +10,7 @@ from xngin.stats.balance import (
     preprocess_for_balance_and_stratification,
     restore_original_numeric_columns,
 )
+from xngin.stats.stats_errors import StatsError
 
 STOCHATREAT_STRATUM_ID_NAME = "stratum_id"
 STOCHATREAT_TREAT_NAME = "treat"
@@ -72,6 +73,9 @@ def assign_treatment_and_check_balance(
     # didn't originally recognize when creating the dataframe. This might have arisen from nullable
     # numeric columns in the underlying database being converted to object types.
     orig_data_to_stratify = df[[id_col, *orig_stratum_cols]].infer_objects()
+    # Protect against inadvertent precision loss from stochatreat silently upcasting your index.
+    orig_data_to_stratify[id_col] = orig_data_to_stratify[id_col].astype("object")
+
     df_cleaned, exclude_cols_set, numeric_notnull_set = (
         preprocess_for_balance_and_stratification(
             data=orig_data_to_stratify,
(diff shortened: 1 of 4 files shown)


src/xngin/stats/assignment.py
```diff
@@ -72,6 +73,9 @@ def assign_treatment_and_check_balance(
     # didn't originally recognize when creating the dataframe. This might have arisen from nullable
     # numeric columns in the underlying database being converted to object types.
     orig_data_to_stratify = df[[id_col, *orig_stratum_cols]].infer_objects()
+    # Protect against inadvertent precision loss from stochatreat silently upcasting your index.
+    orig_data_to_stratify[id_col] = orig_data_to_stratify[id_col].astype("object")
```
poornimaramesh, line 77: Wouldn't it be better to cast the ID column to "string" instead of "object"?
This way we're explicitly typesetting the column and preventing the rounding errors (unless we're using "object" to handle nulls also?)

src/xngin/stats/assignment.py
```diff
@@ -105,10 +109,16 @@ def assign_treatment_and_check_balance(
...
     )
+
     df_cleaned = df_cleaned.merge(treatment_status, on=id_col)
     stratum_ids = df_cleaned[STOCHATREAT_STRATUM_ID_NAME]
     treatment_ids = df_cleaned[STOCHATREAT_TREAT_NAME]
 
+    if len(df) != len(treatment_ids):
+        raise StatsError(
+            f"Expected {len(df)} assignments but only have {len(df_cleaned)}"
+        )
```
poornimaramesh, line 120: given the error message, should we not be checking if len(df) != len(df_cleaned)?

src/xngin/stats/assignment.py
qixotic: Object should preserve the bit representation; we don't want to inadvertently mutate anything. (Theoretically, one could use a float even as the participant_id if you knew it was unique in your case, and those might not have an exact string representation.) 

This is also an internal transform to a temp df for stochatreat to use. (I want to push this fix into that lib one day as well where it ideally should be.)

src/xngin/stats/assignment.py
qixotic: Sure that would be less confusing. :)

src/xngin/stats/test_assignment.py
```diff
@@ -287,3 +287,22 @@ def test_simple_random_assignment_with_different_seeds(sample_df):
...
+    sample_df.loc[2, "id"] = 103241243500726324
+    result = assign_treatment_and_check_balance(
+        df=sample_df,
+        stratum_cols=["gender", "region"],
+        id_col="id",
+        n_arms=2,
+        random_state=42,
+    )
+
+    assert len(result.treatment_ids) == len(sample_df)
```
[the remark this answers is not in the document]
qixotic: Added a better description, thanks.

This particular test can't verify the set of final ids are the same, since that association isn't returned. (We could certainly have echoed `df_cleaned[id_col]` back though in the AssignmentResult struct though; not necessary but may have made debugging easier, and something to do if yet another bug arises!)

PR #55, opened 2025-08-22
Remove obsolete isinstance(..., ParticipantsDef) checks from admin_api.

dctaf:


qixotic (approved):
Ah, this catches one more instance that 54 didn't.

PR #61, opened 2025-08-27
Snapshots API and scaffolding for periodic snapshot jobs

dctaf:
Shaves many yaks in the support of implementing experiment snapshots.

1. Adds a CRUD-like Admin API for requesting, listing, and managing experiment snapshots.
2. Introduces the "snapshotter" tool for capturing snapshots. Intended to be invoked periodically as a cronjob.

There is placeholder for implementing the actual snapshot collection
functionality; the tables.Snapshots.data field and
`snapshotter._query_dwh_for_snapshot_data` method both use a `dict` as a
placeholder.

### DIFF
--- a/src/xngin/apiserver/routers/admin/admin_api.py
+++ b/src/xngin/apiserver/routers/admin/admin_api.py
@@ -70,7 +70,6 @@ from xngin.apiserver.routers.admin.admin_api_types import (
     CreateSnapshotResponse,
     DatasourceSummary,
     EventSummary,
-    ExperimentPathParams,
     GetDatasourceResponse,
     GetOrganizationResponse,
     GetSnapshotResponse,
@@ -87,7 +86,6 @@ from xngin.apiserver.routers.admin.admin_api_types import (
     OrganizationSummary,
     PostgresDsn,
     RedshiftDsn,
-    SnapshotPathParams,
     SnapshotStatus,
     UpdateDatasourceRequest,
     UpdateOrganizationRequest,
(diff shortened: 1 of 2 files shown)


src/xngin/apiserver/routers/admin/admin_api.py
```diff
@@ -314,6 +329,22 @@ async def validate_webhooks(
...
+class DatasourcePathParams(OrganizationPathParams):
+    datasource_id: Annotated[str, Path()]
+
+
+class ExperimentPathParams(DatasourcePathParams):
+    experiment_id: Annotated[str, Path()]
+
+
+class SnapshotPathParams(ExperimentPathParams):
+    snapshot_id: Annotated[str, Path()]
```
poornimaramesh, line 345: can we move these to `admin_api_types.py`?

src/xngin/apiserver/snapshots/snapshotter.py
```diff
@@ -0,0 +1,135 @@
...
+            snapshot.data = await _query_dwh_for_snapshot_data(session, datasource, experiment)
+            snapshot.status = "success"
+    except TimeoutError:
+        logger.info(f"{experiment.id}.{snapshot.id}: timeout")
+        snapshot.status = "failed"
+        snapshot.message = "TimeoutError"
+    except Exception as exc:
+        logger.opt(exception=exc).info(f"{experiment.id}.{snapshot.id}: unhandled exception")
+        snapshot.status = "failed"
+        snapshot.message = f"{type(exc).__name__}: {exc}"
```
poornimaramesh, line 123: we don't seem to be handling timeout errors differently than some other error. Is this a placeholder for us to implement differential error handling? Could we include a #TODO if so?

PR #70, opened 2025-09-17
Fix balance check's categorical coding

qixotic:
## Description

Fixes our use of pd.dummies for generating indicator variables out of categoricals by using patsy's support for coding categorical data. Our old approach would naively embed strings into the dummy column name then join them together to make the final formula, which could break if the strings had improper characters.

## How has this been tested?

unittests

## Checklist

Fill with `x` for completed.

### DIFF
--- a/src/xngin/stats/balance.py
+++ b/src/xngin/stats/balance.py
@@ -139,16 +139,9 @@ def check_balance_of_preprocessed_df(
     # Convert all non-numeric columns into dummy vars, including booleans
     non_numeric_columns = {c for c in data.columns if not is_any_real_numeric_dtype(data[c])}
     cols_to_dummies = list(non_numeric_columns - exclude_col_set)
-    df_analysis = pd.get_dummies(
-        data,
-        columns=cols_to_dummies,
-        prefix=cols_to_dummies,
-        dummy_na=False,
-        drop_first=True,
-    )
 
     # Create formula excluding specified columns
-    covariates = [col for col in df_analysis.columns if col != treatment_col and col not in exclude_col_set]
+    covariates = [col for col in data.columns if col != treatment_col and col not in exclude_col_set]
     if len(covariates) == 0:
         raise StatsBalanceError(
             "No usable fields for performing a balance check found. Please check your metrics "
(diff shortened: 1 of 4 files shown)


src/xngin/stats/balance.py
```diff
@@ -139,16 +139,9 @@ def check_balance_of_preprocessed_df(
...
-        data,
-        columns=cols_to_dummies,
-        prefix=cols_to_dummies,
-        dummy_na=False,
-        drop_first=True,
-    )
 
     # Create formula excluding specified columns
-    covariates = [col for col in df_analysis.columns if col != treatment_col and col not in exclude_col_set]
+    covariates = [col for col in data.columns if col != treatment_col and col not in exclude_col_set]
```
poornimaramesh, line 144: just nitpicking, but wouldn't it be easier to write the condition as `col not in exclude_col_set + [treatment_col]`?

src/xngin/stats/balance.py
```diff
@@ -157,9 +150,10 @@ def check_balance_of_preprocessed_df(
 
     # TODO(roboton): Run multi-class regression via MVLogit
     # df_analysis[treatment_col] = pd.Categorical(df_analysis[treatment_col])
```
poornimaramesh, line 152: can we take this line out entirely? seems that it's been commented out for a while...

src/xngin/stats/test_assignment.py
```diff
@@ -173,11 +173,11 @@ def test_assign_treatment_with_infer_objects():
...
 
 
-def test_assign_treatment_decimal_strata_columns_are_not_supported(sample_df):
-    """Test that unconverted Decimal strata columns raise an error."""
+def test_assign_treatment_decimal_strata_columns_does_not_raise_error(sample_df):
+    """Test that unconverted Decimal strata columns does NOT raise an error."""
     sample_df["decimal"] = sample_df["income"].apply(Decimal)
-    with pytest.raises(RecursionError):
-        assign_treatment_and_check_balance(df=sample_df, stratum_cols=["decimal"], id_col="id", n_arms=2)
+    # Ok since we treat it as a categorical (for better or worse).
```
poornimaramesh, line 179: is this something we might have to address in future? Could we create an issue to track it, if so?

poornimaramesh (approved):
Looks good, tests pass. Thanks!

src/xngin/stats/test_assignment.py
qixotic: Our consumer code around the stats lib does the conversion here: https://github.com/agency-fund/evidential-be/blob/b0bda33381fd01e01ff673e517b043b69c669a7d/src/xngin/apiserver/routers/assignment_adapters.py#L76 

So we're good 👍

PR #77, opened 2025-09-22
Bandit snapshots

poornimaramesh:
## Ticket

https://github.com/agency-fund/evidential-sprint/issues/137

## Description
Update snapshots to also include bandit experiments

## How has this been tested?
- Manually tested snapshot endpoints with bandit experiments
- `task start-snapshotter` after creating bandit experiments
- Run unit-tests to check nothing breaks


src/xngin/apiserver/snapshots/snapshotter.py
```diff
@@ -143,15 +149,39 @@ async def _query_dwh_for_snapshot_data(
...
+
+            if len(draws) == 0:
+                context_vals = [0.0] * len(contexts)
+            else:
+                sorted_contexts = sorted(contexts, key=lambda c: c.id)
+                all_context_vals = [draw.context_vals for draw in draws if draw.context_vals is not None]
+                mean_context_val = np.mean(all_context_vals, axis=0)
+                rng = np.random.default_rng(RANDOM_STATE)
+                context_vals = [
+                    rng.binomial(1, m) if context.value_type == ContextType.BINARY else m
```
qixotic, line 171: Why do you sample from the binomial distribution if it was a binary context, rather than just using the mean here as well? Wouldn't the sampling just result in a more noisy analysis?

poornimaramesh: That's a good point. The context input is expected to be binary valued, and averaging will give us a value between 0 and 1. I've switched to rounding up to 1 for values greater than 0.5 and rounding down to 0. otherwise. That should make it less noisy (and also solves the random state problem you pointed out above)

The proper way to do this is to marginalize over all possible context inputs, but it requires changing other parts of the code -_-

src/xngin/apiserver/snapshots/snapshotter.py
```diff
@@ -143,15 +149,39 @@ async def _query_dwh_for_snapshot_data(
...
+        # the mean context values. Captured in issue 140
+        # (https://github.com/agency-fund/evidential-sprint/issues/140)
+        if ExperimentsType(experiment.experiment_type).is_cmab():
+            draws = await experiment.awaitable_attrs.draws
+            contexts = await experiment.awaitable_attrs.contexts
+
+            if len(draws) == 0:
+                context_vals = [0.0] * len(contexts)
+            else:
+                sorted_contexts = sorted(contexts, key=lambda c: c.id)
```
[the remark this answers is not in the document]
poornimaramesh: updated, thanks for catching

PR #81, opened 2025-09-30
update_experiment now rejects invalid URLs.

dctaf:
This fixes the current issue where experiment design URLs can be updated to invalid URLs which break downstream validation. We retain special treatment of NULLs to support our ad-hoc "patch" protocol where "None" means "no change", and empty string means "clear".

### DIFF
--- a/src/xngin/apiserver/routers/admin/test_admin.py
+++ b/src/xngin/apiserver/routers/admin/test_admin.py
@@ -1478,6 +1478,27 @@ async def test_update_experiment_url_valid(testing_experiment, ppatch, pget, url
     assert parsed_response.design_spec.design_url.encoded_string() == expected_url
 
 
+async def test_update_experiment_url_null_when_empty(testing_experiment, ppatch, pget):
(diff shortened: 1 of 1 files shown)


src/xngin/apiserver/routers/admin/test_admin.py
```diff
@@ -1447,6 +1447,37 @@ async def test_update_experiment(testing_experiment, ppatch, pget):
...
+
+@pytest.mark.parametrize(
+    ("url", "expected_url"),
+    [
+        ("http://example.com", "http://example.com/"),
+        ("https://example.com", "https://example.com/"),
+        ("https://drive.google.com/...?q=1", "https://drive.google.com/...?q=1"),
+    ],
+)
+async def test_update_experiment_url_valid(testing_experiment, ppatch, pget, url, expected_url):
```
qixotic, line 1468: Can you add one more test in which a known non-empty design url is then updated with a "", resulting in the next fetch of the experiment to have `design_url is None`?

PR #83, opened 2025-10-13
Add middleware to allow valid requests to be embedded within arbitrary request bodies.

dctaf:
Allow API request bodies to be embedded within request bodies we do not
control. The path to the actual request body is specified as a URL parameter
`?_unwrap=`.

In this example, the request handler will receive {"actual": "content"}.

```
POST /api/endpoint?_unwrap=/data/payload
Content-Type: application/json
{"data": {"payload": {"actual": "content"}}}
```

Note that the `_unwrap=` parameter, if present, must be a valid [JSONPointer](https://jg-rp.github.io/python-jsonpath/pointers/) path so they normally will start with a `/`.

### DIFF
--- a/src/xngin/apiserver/test_request_encapsulation_middleware.py
+++ b/src/xngin/apiserver/test_request_encapsulation_middleware.py
@@ -41,7 +41,7 @@ def create_test_app(unwrap_param: str = "_unwrap") -> Starlette:
 @pytest.fixture(scope="module")
 def client():
     app = create_test_app()
-    return TestClient(app)
+    return TestClient(app, raise_server_exceptions=False)
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


src/xngin/apiserver/test_request_encapsulation_middleware.py
```diff
@@ -0,0 +1,298 @@
...
+
+def test_handles_empty_request_body(client):
+    response = client.post(
+        "/echo?_unwrap=/data",
+        data=None,
+        headers={"Content-Type": "application/json"},
+    )
+
+    assert response.status_code == 200
+    assert response.json() == {"received": None}
```
qixotic: So this is a special case in which we ignore the invalid JSONPointer path (since it can't point to anything valid inside an empty body) and call it a success regardless of what _unwrap is set to. I see on line 94 of the middleware where we explicitly return when the body is empty. Just calling out this different behavior from `test_error_on_invalid_paths` due to the body.

qixotic (approved):

PR #84, opened 2025-10-20
Minor fixes to generated OpenAPI spec.

dctaf:
(1) Adds OpenAPI documentation to the authentication headers and move datasource_dependency closer to where it is used.

(2) This makes minor changes to the return types for certain operations: by explicitly setting
`status_code=204`, the generated responses won't have a type definition for 200,
which makes the generated downstream code cleaner.

There is no behavior change.

Here is the generated Orval code before this change (note return type):

```typescript
export const abandonExperiment = async (
  datasourceId: string,
  experimentId: string,
  options?: RequestInit,
): Promise<unknown | void> => {
  return orvalFetch<unknown | void>(
    getAbandonExperimentUrl(datasourceId, experimentId),
    {
      ...options,
      method: "POST",
    },
  );
};
```

Here is the method afterwards:

```typescript
export const abandonExperiment = async (
  datasourceId: string,
  experimentId: string,
  options?: RequestInit,
): Promise<void> => {
  return orvalFetch<void>(getAbandonExperimentUrl(datasourceId, experimentId), {
    ...options,
    method: "POST",
  });
};
```

The consequences on the generated Rust code are more significant.

### DIFF
--- a/src/xngin/apiserver/routers/auth/auth_dependencies.py
+++ b/src/xngin/apiserver/routers/auth/auth_dependencies.py
@@ -135,7 +135,10 @@ def session_token_crypter_dependency():
 
 
 async def require_valid_session_token(
-    authorization: Annotated[HTTPAuthorizationCredentials, Depends(HTTPBearer())],
+    authorization: Annotated[
+        HTTPAuthorizationCredentials,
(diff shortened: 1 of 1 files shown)


dctaf:
The tests are failing because Docker Hub is having problems. Unit tests pass locally.

PR #92, opened 2025-10-25
add endpoint to filter online assignment requests server-side

qixotic:
## Description

Adds an endpoint for frequentist online assignments that support filtering participants on the server side given a list of properties before issuing an assignment.

Also adds common value validation across Filter values at experiment creation, participant querying, and assignment fetching for online experiments that do server-side filtering. (See `validate_filter_value`)

Minor corresponding FE PR: https://github.com/agency-fund/evidential-fe/pull/160

## How has this been tested?

unittests, including testing against our bq instance

## Checklist

Fill with `x` for completed.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -164,7 +164,7 @@ jobs:
           uv run xngin-cli create-apiserver-db
 
   bq-integration:
-    timeout-minutes: 5
+    timeout-minutes: 10
     runs-on: ubuntu-22.04
     needs: changes
     if: needs.changes.outputs.queries == 'true'


src/xngin/apiserver/routers/experiments/experiments_api.py
```diff
@@ -139,6 +141,47 @@ async def get_assignment_for_participant_with_apikey(
         experiment=experiment,
         participant_id=participant_id,
         create_if_none=create_if_none,
+        properties=None,
+        random_state=random_state,
+    )
+
+
+@router.post(
+    "/experiments/{experiment_id}/assignments/{participant_id}/assign_with_filters",
```
qixotic: @poorva-noora this is the proposed endpoint for online assignments; please also see OnlineAssignmentWithFiltersRequest. Should be pretty straightforward, but let me know if you have concerns about the simple payload to POST, thanks!

src/xngin/apiserver/routers/experiments/experiments_api.py
```diff
@@ -139,6 +141,47 @@ async def get_assignment_for_participant_with_apikey(
...
+    description="""
+    Get or create a frequentist online arm assignment for a participant that requires server-side
+    filtering. If an assignment already exists, the properties in the
+    OnlineAssignmentWithFiltersRequest are ignored and the existing assignment is returned.""",
+)
+async def get_online_experiment_assignment_for_participant(
```
dctaf: Nit: consider improving the naming consistency of some of these endpoints..

| Current | Suggestion | Why |
| --- | --- | --- |
| get_assignment_for_participant_with_apikey | get_assignment | API keys are not a feature-level concern. "Participant" is redundant because "assignment" is singular and obvious because we only make assignments for participants. |
| get_online_experiment_assignment_for_participant | get_assignment_filtered | Perhaps we want the filtered approach to apply to non-online experiments in the future. |
| get_cmab_experiment_assignment_for_participant | get_assignment_cmab | This is the same intention as get_assignment_for_participant, but specifically for CMAB. |

src/xngin/apiserver/routers/experiments/property_filters.py
```diff
@@ -0,0 +1,104 @@
...
+        if not _passes_filter(f, field_type, props.get(f.field_name)):
+            return False
+
+    return True
+
+
+def _passes_filter(exp_filter: Filter, field_type: DataType, value: PropertyValueTypes) -> bool:
+    """Check that a value passes a filter."""
+    py_value = _validate_value(value, field_type)
+    parsed_values = [_validate_value(v, field_type) for v in exp_filter.value]
```
dctaf: My understanding is that validating the values in Filter should only be necessary if the existing invariants on Filter are weak. Otherwise, we run the risk of introducing tighter constraints when it is used for online filtering vs when it is used for filtering from the data warehouse. The more we can ensure that this implementation can match the filtering behavior of the SQLAlchemy-generating behavior, the happier we will be long term.

src/xngin/apiserver/routers/experiments/property_filters.py
```diff
@@ -0,0 +1,104 @@
...
+
+    match exp_filter.relation:
+        case Relation.INCLUDES:
+            return py_value in parsed_values
+        case Relation.EXCLUDES:
+            return py_value not in parsed_values
```
dctaf: Do we need special handling for None/null?

src/xngin/apiserver/routers/experiments/test_property_filters.py
```diff
@@ -0,0 +1,505 @@
...
+
+import pytest
+
+from xngin.apiserver.routers.common_api_types import DataType, Filter, PropertyValueTypes
+from xngin.apiserver.routers.common_enums import Relation
+from xngin.apiserver.routers.experiments.property_filters import passes_filters
+
+
+@dataclass
+class Case:
```
dctaf: This is a great approach. It would be helpful to run the SQLAlchemy-backed filtering logic through the same tests so that we know our filtering implementation is similar to how it is evaluated in a SQL context. What do you think about running these tests against the filters in queries.py  too?

src/xngin/apiserver/routers/common_api_types.py
```diff
@@ -140,6 +141,33 @@ def check_has_only_one_of_pct_change_or_target(self) -> Self:
...
+class OnlineAssignmentWithFiltersRequest(ApiBaseModel):
+    """Request model for creating a new frequentist online assignment with server-side filtering.
+
+    Usage notes:
+    1. Currently only used for FREQ_ONLINE experiments.
+    2. If the experiment defines no filters, use the corresponding GET endpoint instead.
+    3. If an assignment already exists for a given participant, the property list is ignored and assignment returned.
+    4. Property names must reference a valid field_name from the participant_type defined in the experiment.
+    5. Property list may be empty.
+    6. If a filter is specified but no value is found, it is treated as NULL.
```
dctaf: If we anticipate that this filtering behavior will be different than the SQL-backed filtering behavior, we should mention that there are differences here.

src/xngin/apiserver/routers/experiments/test_property_filters.py
qixotic: Great idea; I was originally inspired by queries.py, so not a stretch to do what you suggested. I've made that change, although note that bq tests run veeeery slowly, so these new cases add another 10 min almost. :(

src/xngin/apiserver/dwh/test_queries.py
```diff
@@ -176,18 +178,18 @@ def __str__(self):
         return " and ".join([f"{f.field_name} {f.relation.name} {f.value}" for f in self.filters])
 
 
-@pytest.fixture(name="queries_session")
-def fixture_queries_session():
-    """Yields a session that tests can use to operate on a testing database.
+@pytest.fixture(scope="module")
+def queries_dwh_engine():
+    """Yields a SQLAlchemy Engine for tests to ooperate on a test data warehouse.
```
dctaf:
```suggestion
    """Yields a SQLAlchemy Engine for tests to operate on a test data warehouse.
```

src/xngin/apiserver/dwh/test_queries.py
```diff
@@ -176,18 +178,18 @@ def __str__(self):
...
 
-    The testing data warehouse is specified by the XNGIN_QUERIES_TEST_URI environment variable. Usually this is a local
-    Postgres instance, but some integration tests may point these tests at a BigQuery dataset (see CI for example).
+    This dwh is specified by the XNGIN_QUERIES_TEST_URI environment variable. Usually this is a
+    local Postgres, but some integration tests may point at a BigQuery dataset (see CI for example).
```
dctaf:
```suggestion
    This dwh is specified by the XNGIN_QUERIES_TEST_URI environment variable. Usually this is a
    local Postgres, but some integration tests may point at an existing BigQuery dataset (see CI for example).
```

src/xngin/apiserver/dwh/test_queries.py
```diff
@@ -206,49 +208,56 @@ def fixture_queries_session():
...
-            finally:
-                session.close()
+        yield engine
     finally:
-        Base.metadata.drop_all(engine)
         engine.dispose()
 
 
+@pytest.fixture(scope="module")
+def shared_sample_tables(queries_dwh_engine):
```
dctaf: At some point previously we adopted a pattern of prefixing fixture methods with `fixture_x` and using  `@fixture(name="x")` to make them more ergonomic to use. I am not attached to that convention but because we use it everywhere else right now, we might as well stick to it?

dctaf (approved):

PR #95, opened 2025-10-27
Some upgrades required for Python 3.14 compatibility.

dctaf:
These updates are in anticipation of compatibility with Python 3.14. numpy, and possibly others, will be handled in later PRs.


dctaf:
Related: https://github.com/agency-fund/evidential-sprint/issues/188

@qixotic this PR isn't terribly important right now so this is low priority but it would be nice to get ready for 3.14!

PR #106, opened 2025-11-12
Fix analysis edge cases

qixotic:
## Description

Handle analysis edge cases:
* no fetched rows for analysis (we've seen this happen when the wrong ids were used)
* no fetched rows for a subset of arms (e.g. an Online experiment just started and only has a few arms with non-zero assignments so far)
* fetched rows for _all_ arms for a particular metric are all missing values (can happen for one-time-only metrics or metrics that were newly constructed in your dwh)

Adds a new `testing_dwh.csv.zst` column: `is_onboarded_onetime` simulating a "one-time metric", with the first 500k rows having the value NULL, and the remaining being a copy of `is_onboarded`.

Also add myself to stats CODEOWNERS so I don't need to always bug Poornima for non-stats changes to that module. :)

Corresponding FE PR: https://github.com/agency-fund/evidential-fe/pull/166

## How has this been tested?

unittests

Also query prod db to verify that FreqExperimentAnalysisResponse.metric_analyses all have non-null metric_name and metric fields, and that the names match the nested field_name:
```
select 
    ma.value->>'metric_name', ma.value->'metric'->>'field_name', * 
from snapshots s
CROSS JOIN LATERAL jsonb_array_elements(s.data->'metric_analyses') AS ma(value)
where
    (
        ma.value->>'metric' is NULL
        OR ma.value->>'metric_name' is NULL
        OR ma.value->>'metric_name' != ma.value->'metric'->>'field_name'
    )
    AND s.data->>'type' not in ('bandit')
    AND s.data IS NOT NULL
;
```

Also extended the test dwh data to be able to write a unittest for one edge case. DuckDB CLI was used to add a new column (thanks @dctaf for the reference in our old xngin repo):

```
CREATE TABLE dwh AS SELECT * FROM 'src/xngin/apiserver/testdata/testing_dwh.csv.zst';  -- duckdb syntax
select is_onboarded, count(*) from dwh where id <= 500000 group by 1;
┌──────────────┬──────────────┐
│ is_onboarded │ count_star() │
│   boolean    │    int64     │
├──────────────┼──────────────┤
│ false        │       349825 │
│ true         │       150175 │
└──────────────┴──────────────┘
select is_onboarded, count(*) from dwh where id > 500000 group by 1;
┌──────────────┬──────────────┐
│ is_onboarded │ count_star() │
│   boolean    │    int64     │
├──────────────┼──────────────┤
│ true         │       149966 │
│ false        │       350034 │
└──────────────┴──────────────┘

ALTER TABLE dwh ADD is_onboarded_onetime BOOLEAN NULL;

WITH maxid AS (
    SELECT max(id) AS max_id FROM dwh
)
-- SELECT * from maxid;
-- next CTE to hold our new column values
, newdata AS (  
    SELECT
        dwh.id,
        -- a copy of is_onboarded, but with the first half NULLed out
        CASE
            WHEN dwh.id <= maxid.max_id / 2 THEN NULL
            ELSE dwh.is_onboarded
        END AS is_onboarded_onetime
    FROM dwh CROSS JOIN maxid 
)
-- select * from newdata where is_onboarded_onetime is NOT NULL;
UPDATE dwh
SET is_onboarded_onetime = newdata.is_onboarded_onetime
FROM newdata WHERE dwh.id = newdata.id;

SELECT is_onboarded_onetime, count(*) from dwh group by 1;
┌──────────────────────┬──────────────┐
│ is_onboarded_onetime │ count_star() │
│       boolean        │    int64     │
├──────────────────────┼──────────────┤
│ false                │       350034 │
│ NULL                 │       500000 │
│ true                 │       149966 │
└──────────────────────┴──────────────┘

COPY dwh (
    gender,
    ethnicity,
    first_name,
    last_name,
    id,
    income,
    potential_0,
    potential_1,
    is_recruited,
    is_registered,
    is_onboarded,
    is_onboarded_onetime,
    is_engaged,
    is_retained,
    baseline_income,
    current_income,
    uuid_filter,
    sample_date,
    sample_timestamp,
    timestamp_with_tz)
TO 'src/xngin/apiserver/testdata/testing_dwh.csv.zst';
```

And then finally, we recompress the zst file before adding it to the commit:
```
ls -lh src/xngin/apiserver/testdata/testing_dwh.csv.zst
zstd -d -c src/xngin/apiserver/testdata/testing_dwh.csv.zst | zstd -9 > /tmp/smaller.tgz
mv /tmp/smaller.tgz src/xngin/apiserver/testdata/testing_dwh.csv.zst
ls -lh src/xngin/apiserver/testdata/testing_dwh.csv.zst
```

## Checklist

Fill with `x` for completed.


dctaf (approved):
The changes to common_api_types suggest a `task genapiclient` run is worthwhile.

qixotic:
> The changes to common_api_types suggest a `task genapiclient` run is worthwhile.

Yes, opened the corresponding PR 👍

src/xngin/apiserver/routers/experiments/experiments_common.py
```diff
@@ -1022,20 +1029,36 @@ async def analyze_experiment_freq_impl(
...
-                    arm_id=arm.id,
-                    arm_name=arm.name,
-                    arm_description=arm.description,
-                    is_baseline=arm_result.is_baseline,
-                    estimate=arm_result.estimate,
-                    p_value=arm_result.p_value,
-                    t_stat=arm_result.t_stat,
-                    std_error=arm_result.std_error,
-                    num_missing_values=arm_result.num_missing_values,
+            if arm.id in analyze_results[metric_name]:
```
poornimaramesh: Is it possible that all arms have missing values for a given metric in every row, even if participant IDs are configured correctly in the dwh? L1032 would fail in this case no, because `analyze_freq_experiment` would return an empty dictionary? I didn't see a test for this, so I are we sure this won't happen?

qixotic: Yes, a 'one-time-metric' could have all rows be NA initially until the outcome is finally recorded in the dwh. There's an else condition at 1049 to handle when we don't find the arm. (The FE PR shows the corresponding UI when num_missing_values=-1

src/xngin/stats/test_analysis.py
```diff
@@ -167,3 +168,39 @@ def test_analysis_with_missing_outcomes(test_assignments, test_outcomes):
...
+        else:
+            assert metric_results[arm_id].estimate is not None
+            assert metric_results[arm_id].p_value is not None
+            assert metric_results[arm_id].t_stat is not None
+            assert metric_results[arm_id].std_error is not None
+            assert metric_results[arm_id].num_missing_values == 0
+
+    # But when *all* arms are missing values, we should get an empty result,
+    # since no regression was performed.
+    num_missing_values = len(test_outcomes)
```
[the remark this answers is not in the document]
qixotic: See above; we should be covered when no analysis is generated for an arm!

PR #113, opened 2025-11-18
Add backend support for unbalanced arms in frequentist experiments.

qixotic:
## Description

Add support for an array of `arm_weights` (floats that must sum up to 100) that can be included when designing a frequentist experiment to allow for unbalanced arm allocations, both in pre-assigned and online frequentist experiments.

## Related PRs

Frontend: https://github.com/agency-fund/evidential-fe/pull/172

## Future Tasks

* Update stochatreat when https://github.com/manmartgarc/stochatreat/pull/52 is resolved.

## How has this been tested?

unittests, including cross-checking against other web-based power calculators

## Checklist

Fill with `x` for completed.


src/xngin/apiserver/routers/experiments/experiments_common.py
```diff
@@ -758,11 +759,25 @@ async def create_assignment_for_participant(
...
+                # Weighted random selection. First convert to probabilities.
+                sum_weights = sum(experiment.arm_weights)
+                weights = [w / sum_weights for w in experiment.arm_weights]
+                # Get weights in the same order as sorted_arms.
+                arm_id_to_weight = {arm.id: w for arm, w in zip(experiment.arms, weights, strict=True)}
+                arm_weights_sorted = [arm_id_to_weight[arm.id] for arm in sorted_arms]
+
+            rng = np.random.default_rng(random_state)
+            index = rng.choice(len(sorted_arms), p=arm_weights_sorted)
+            chosen_arm = sorted_arms[index]
```
dctaf: create_assignment_for_participant is ~120 lines of validation, database reads and writes, and statistical business logic. Readability could be improved by keeping our methods at consistent levels of detail; so, consider extracting the "equal allocation - use a simple random choice" behavior to `choose_online_arm(...)`, similar to how the case below invokes a helper method. That would allow it to be tested in isolation.

Later, we could move this towards a side-effect free method -- instead of mutating the database directly, it could return instructions for the caller describing changes that should be made persistent. This would be easier to test and benchmark.

poornimaramesh (approved):

PR #129, opened 2025-12-29
Bootstrap developer accounts with freq_{online,preassigned} experiments.

dctaf:
Various entities are created when a developer creates a new account, such as a datasource corresponding to the local testing dwh. This PR adds one preassigned and one frequentist experiment.

### DIFF
--- a/src/xngin/apiserver/middleware.py
+++ b/src/xngin/apiserver/middleware.py
@@ -13,4 +13,4 @@ def setup(app):
         allow_origins=["*"],
         max_age=7200,  # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age
     )
-    app.add_middleware(RequestEncapsulationMiddleware)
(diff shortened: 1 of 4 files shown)


src/xngin/apiserver/storage/bootstrap.py
```diff
@@ -1,51 +1,112 @@
...
+    session: AsyncSession, user: tables.User, testing_dwh_dsn: str | None
+) -> tables.User:
     """Adds models to User such that they can have a good first time experience with the application.
 
-    Users will have an organization and a NoDWH datasource created for them. If dsn is provided, we create that
-    datasource and a participant type for them (assuming DSN refers to a testing_dwh instance).
+    Users will have an organization and a NoDWH datasource created for them.
+
+    If testing_dwh_dsn is provided, a datasource and a participant type for that DWH will be created. testing_dwh_dsn
+    must refer to a testing dwh instance. This is usually only used in development environments.
```
poornimaramesh: can we update the docs to say we're also adding test experiments?

PR #130, opened 2025-12-30
add error handling for db connections

poornimaramesh:
## Ticket
Fixes issue [234](https://github.com/agency-fund/evidential-sprint/issues/234)

## Related PRs

Frontend: [PR 191](https://github.com/agency-fund/evidential-fe/pull/191)

## Description
Better error handling for database errors
We already handle DB not found errors, and we have config checking but don't surface those errors to the user. So I
- Added error handling when trying to create an db engine session (required for BQ)
- Added error handling when trying to inspect db (required for Postgres and Redshift)
I catch the errors to create a standard "connection error", append the error details to it, and return it through the `inspect` endpoint.

This seemed like the simplest fix, but the error message isn't always informative -- we should think about how to improve this (through the UI?)

## How has this been tested?
- Added a test for this for each of the allowed DB types.

## Checklist

Fill with `x` for completed.

### DIFF
--- a/src/xngin/apiserver/routers/admin/test_admin.py
+++ b/src/xngin/apiserver/routers/admin/test_admin.py
@@ -781,7 +781,7 @@ def test_datasource_errors(pget, ppost):
             dsn=PostgresDsn(
                 host="127.0.0.1",
                 user="postgres",
-                port=5432,
+                port=5499,
                 password=RevealedStr(value="postgres"),
                 dbname="nonexistent_db",
                 sslmode="disable",


src/xngin/apiserver/routers/admin/admin_api.py
```diff
@@ -961,12 +962,14 @@ async def inspect_datasource(
...
                 tablenames = await dwh.list_tables()
-
             ds.set_table_list(tablenames)
             await session.commit()
             return InspectDatasourceResponse(tables=tablenames)
+
         except DwhDatabaseDoesNotExistError as exc:
             raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=str(exc)) from exc
+        except DwhConnectionError as exc:
+            raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=str(exc)) from exc
```
dctaf: 1. I recommend adding this (and DwhDatabaseDoesNotExistError) to exceptionhandlers.py so that we get the same behavior across all endpoints that interact with the data warehouse.
2. For DwhConnectionError: 
     * HTTP status codes don't map cleanly onto all application error conditions but some are more appropriate than others.  We should choose one that indicates an upstream error.
     * HTTP 4xx codes are intended as "client did something wrong" -- 400 specifically means that the request is invalid and that sending it again will also return a 400. That's not quite what we want to do here because a subsequent request might succeed.
     * Suggestion: Let's use a 502 instead -- this signals that we are acting as a proxy of some kind, and our upstream connection failed.

dctaf (changes requested):

src/xngin/apiserver/routers/admin/admin_api.py
poornimaramesh: Done, thanks!

src/xngin/apiserver/routers/admin/admin_api.py
```diff
@@ -183,6 +182,17 @@ async def lifespan(_app: FastAPI):
     yield
 
 
+@asynccontextmanager
+async def clear_db_table_cache_on_error(session: AsyncSession, datasource: tables.Datasource):
+    """Context manager that clears a datasource's cached table list on error."""
+    try:
+        yield
+    except:
```
dctaf: As we learn more about the exceptions that are thrown, it will probably be good to constrain this exception clause to DWH-related exceptions. I've learned to be allergic to `except Exception` especially when they have complex side effects such as database operations. This is OK for now though!

PR #133, opened 2026-01-13
Calculate Expected MDE with user-specified sample size

poornimaramesh:
## Ticket

Fixes [issue 107](https://github.com/agency-fund/evidential-sprint/issues/107)

## Related PRs

Frontend: [PR 198](https://github.com/agency-fund/evidential-fe/pull/198)

## Description
We already re-compute percent change (MDE) possible with all available samples when we don't have the required sample size for the desired MDE / power / confidence. I re-use that code to compute observable MDE for the user's chosen_n input.

This includes the following changes:
- minor updates to the calculation code in `stats/power.py` -- I move the percent change computation code to a separate function, and added tests for this function
- updated experiment creation code in `experiments/experiments_common.py` to do the observable MDE computation when the experiment is created, and added tests for this

## How has this been tested?
- Check if all tests pass

## To-do before merge (optional)
I've re-used the calculation that already existed, but please let me know if I should update the power calc (e.g. switch to one-sided instead of two-sided, do additional checks, etc.)

## Checklist

Fill with `x` for completed.


src/xngin/stats/power.py
```diff
@@ -22,6 +22,88 @@ def _power_analysis_error(
...
+        # We always assume the first arm is control.
+        control_prob = weights[0]
+        # Use the largest treatment arm for a conservative estimate.
+        # (larger ratio requires a larger total sample size)
+        max_treatment_prob = max(weights[1:])
+        arm_ratio = max_treatment_prob / control_prob
+
+    control_n_available = int(chosen_n * control_prob)
+
+    if metric.metric_type == MetricType.NUMERIC:
```
dctaf: Optional: If you use a match statement instead of an if, then this could would fail to type check if an additional MetricType were ever introduced due to enum exhaustiveness checking.

src/xngin/stats/power.py
```diff
@@ -22,6 +22,88 @@ def _power_analysis_error(
...
+        # For unbalanced arms, we need to calculate based on the ratio of treatment to control
+        # Convert weights (sum to 100) to probabilities
+        sum_weights = sum(arm_weights)
+        weights = [w / sum_weights for w in arm_weights]
+        # We always assume the first arm is control.
+        control_prob = weights[0]
+        # Use the largest treatment arm for a conservative estimate.
+        # (larger ratio requires a larger total sample size)
+        max_treatment_prob = max(weights[1:])
+        arm_ratio = max_treatment_prob / control_prob
```
dctaf: This duplicates the code at 181-193. Would a helper method be appropriate?

src/xngin/stats/power.py
poornimaramesh: Changed, thanks!

poornimaramesh:
I've rebased on top of the changes from [Muriel's PR](https://github.com/mdulieu/evidential-be/commit/c89b15021c0947dfd4bde45d12f969c5181b8e02) -- the only difference is the power check when we create a pre-assigned experiment now.

The corresponding [frontend PR](https://github.com/agency-fund/evidential-fe/pull/198) has also been updated.

@qixotic would you want to take a quick look before I merge?

PR #137, opened 2026-01-15
Fix unbalanced arms power calculation: use min instead of max

mdulieu:
## Summary
Fixes the power calculation for experiments with unbalanced arm allocation.

## The Bug
The current code uses `max(weights[1:])` to find the treatment arm ratio, but this gives overly optimistic power estimates. The smallest treatment arm has the least power and requires the most total samples to achieve the same statistical power.

## The Fix
Changed from `max()` to `min()` to use the smallest treatment arm for conservative power estimates.

Also updated `test_analyze_metric_power_unbalanced_three_arms` to expect the correct (higher) sample size of 320 instead of 215.

## Example
For an experiment with:
- 50% Control (5,000 people)
- 40% Treatment A (4,000 people)
- 10% Treatment B (1,000 people)

**Before (wrong):** Used 40% arm (4,000 people) for calculations
**After (correct):** Uses 10% arm (1,000 people) for calculations

The 10% arm has the hardest time detecting effects, so we should plan for that.

## Testing
```bash
poetry run pytest src/xngin/stats/test_power.py -v
# Result: 14 passed
```

## Acknowledgment
Bug identified and confirmed by Linus via email discussion.
```

### DIFF
--- a/src/xngin/apiserver/routers/admin/test_admin.py
+++ b/src/xngin/apiserver/routers/admin/test_admin.py
@@ -1386,9 +1386,9 @@ async def test_power_check_with_unbalanced_arms(testing_datasource_with_user, pp
 
     # And again with three arms
     design_spec.arms = [*design_spec.arms, Arm(arm_name="arm3", arm_description="Arm 3")]
-    design_spec.arms[0].arm_weight = 15
-    design_spec.arms[1].arm_weight = 60
-    design_spec.arms[2].arm_weight = 25
+    design_spec.arms[0].arm_weight = 10
+    design_spec.arms[1].arm_weight = 50
+    design_spec.arms[2].arm_weight = 40
     response = ppost(
         f"/v1/m/datasources/{testing_datasource_with_user.ds.id}/power",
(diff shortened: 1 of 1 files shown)


qixotic:
Hi Muriel, I've fixed the issues caught by our automated checks and will merge on your behalf (re: [pre-commit](https://github.com/agency-fund/evidential-be/actions/runs/21022655921/job/60444126496?pr=138), [ruff](https://github.com/agency-fund/evidential-be/actions/runs/21022655921/job/60444126428?pr=138), [unittests](https://github.com/agency-fund/evidential-be/actions/runs/21022655940/job/60444126772?pr=138)).

Please check that you've setup pre-commit (`uv run pre-commit install`) locally to catch these issues on your end as well for future contributions. Thanks!

qixotic:
integration test [failing](https://github.com/agency-fund/evidential-be/actions/runs/21026799631/job/60453453118?pr=137) because of the restricted context not having secrets.GOOGLE_APPLICATION_CREDENTIALS_CONTENT set. Will override merge and fix up our workflow to skip this on forks.

PR #143, opened 2026-01-19
Feature/add desired n mde calculation

mdulieu:
This PR adds the ability to calculate minimum detectable effect (MDE) when a desired sample size is specified, without requiring a target effect size.

## Changes
- Added `desired_n` parameter to `analyze_metric_power()` and `check_power()`
- When `desired_n` is provided (without a metric_target), the function calculates the minimum detectable effect instead of required sample size
- The function now operates in two modes:
  - **Sample size mode** (existing): Given target effect → calculates required N
  - **MDE mode** (new): Given desired N → calculates minimum detectable effect
- Added comprehensive tests for both numeric and binary metrics in MDE mode
- Fixed arm ratio calculation to use `min` weight (merged from main)

## Testing
- All existing tests pass
- Added new tests for MDE calculation mode (without metric_target)
- Verified MDE calculation works for both numeric and binary metrics

### DIFF
--- a/src/xngin/stats/power.py
+++ b/src/xngin/stats/power.py
@@ -109,9 +109,12 @@ def calculate_mde_with_desired_n(
             # Convert Cohen's h back to proportion
             # h = 2 * arcsin(sqrt(p1)) - 2 * arcsin(sqrt(p2))
             # where p1 is baseline and p2 is target
+            # NOTE: typically the target proportion is > baseline, so h will be negative. But when
+            # solving for an effect size given n, it will always be positive, meaning the target
+            # will be *smaller* than baseline.
             p1 = metric.metric_baseline
-            arcsin_p2 = 2 * np.arcsin(np.sqrt(p1)) - min_effect_size
-            target_possible = np.sin(arcsin_p2 / 2) ** 2
+            arcsin_p2 = (2 * np.arcsin(np.sqrt(p1)) - min_effect_size) / 2.0
+            target_possible = np.sin(arcsin_p2) ** 2
         case _:
             raise ValueError(f"metric_type must be one of {list(MetricType)}.")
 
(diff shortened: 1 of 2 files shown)


src/xngin/apiserver/routers/experiments/experiments_common.py
```diff
@@ -137,6 +138,23 @@ async def create_experiment_impl(
...
+            if request.power_analyses:
+                for analysis in request.power_analyses.analyses:
+                    if chosen_n == analysis.target_n:
+                        actual_mde = analysis.metric_spec.metric_pct_change
+                    elif chosen_n == analysis.metric_spec.available_n:
+                        actual_mde = analysis.pct_change_possible
+                    else:
+                        _, actual_mde = calculate_mde_with_desired_n(
+                            metric=analysis.metric_spec,
+                            desired_n=chosen_n,
```
qixotic: Thanks @mdulieu for the merge, although actually I was just referring to bringing over Poorinma's power.py changes. 😅  To keep the changes more logically isolated, could you revert these merged in changes to {test_}experiments_common.py and src/xngin/apiserver/routers/common_api_types.py in your PR, and simply make this one line change so that this code uses your new power.py interface (desired_n name)? 

Then #133 can also become just about the API change to leverage this extended power module functionality.

qixotic (changes requested):

mdulieu:
Refactoring complete. Ready for review.
I've addressed the review feedback:
Merged main and incorporated:

effective_n logic for non-null count handling

Refactored analyze_metric_power() into helper functions:

_analyze_power_mde_mode() for MDE calculation
_analyze_power_sample_size_mode() for sample size calculation
These helpers use calculate_mde_with_desired_n() and _calculate_arm_ratio_and_control_prob_from_weights() from PR #133

Other changes:

Renamed chosen_n → desired_n in power.py
Each helper function validates its own inputs (self-contained)
No more early returns in main function

Reverted per review feedback:

src/xngin/apiserver/routers/common_api_types.py
src/xngin/apiserver/routers/experiments/experiments_common.py
src/xngin/apiserver/routers/experiments/test_experiments_common.py

All tests passing. Pre-commit passing.
Question: I reverted the 3 files to main, but main doesn't have the calculate_mde_with_desired_n call - that's added in PR #133. Should I add just that one block from PR #133 (with desired_n instead of chosen_n), or leave the files as-is (matching main)?

Issue #148, opened 2026-02-03
Extend power API to support desired_n parameter for MDE calculation


mdulieu:
**Background**

PR #143 adds core functionality to calculate minimum detectable effect (MDE) when given a desired sample size, rather than only calculating required sample size from a target effect. This issue proposes extending the backend API to expose this functionality to the frontend UI.

**Motivation**

Users often know their available sample size upfront (due to limited user base, time constraints, or budget) and want to determine what effect size would be detectable with that constraint. This is the reverse of the current flow where users specify an effect size and we calculate the required N.

**Implementation Plan**

Phased approach:

### Phase 1: Core power.py Module (PR #143) - COMPLETED ✓
- Added `desired_n` parameter to `analyze_metric_power()` and `check_power()`
- When provided, calculates MDE instead of required sample size
- Backward compatible, includes comprehensive tests

### Phase 2: Backend API (feature/add-desired-n-api branch) - READY FOR REVIEW

I have implemented these changes:

1. **Add `desired_n` field to `BaseFrequentistDesignSpec`** (common_api_types.py)
   - Optional integer parameter (defaults to `None`)
   - When provided, triggers MDE calculation mode

2. **No changes needed to `power_check()` endpoint** (admin_api.py)
   - Already passes `design_spec.desired_n` to `check_power()`
   - Existing flow automatically supports the new parameter

**Usage Example:**
```json
{
  "design_spec": {
    "experiment_type": "freq_preassigned",
    "desired_n": 5000,
    "metrics": [...],
    // other fields
  }
}
```

### Phase 3: Frontend Integration - FUTURE WORK

After the backend API is working, the frontend would need to:
- Add UI for users to optionally specify `desired_n` during power check
- Display the calculated MDE in the results instead of required sample size
- Handle the workflow for when users want to use this mode

**Question for Review**

As mentioned in PR #143: should we consider doing all power calculations in the power check step, blocking the user from clicking Next until recalculation completes? This could:
- Simplify the overall workflow by consolidating power calculations
- Allow the API to optionally accept `list[DesignSpecMetric]` from the frontend to skip re-querying the data warehouse
- Enable future enhancements like letting users specify custom baseline/stddev values

**Next Steps**

1. Get feedback on this proposal
2. After PR #143 merges, submit PR with API changes from `feature/add-desired-n-api` branch
3. Discuss frontend integration approach separately

mdulieu:
Resolved by PR #143 and PR #149. Frontend integration tracked in #216.

PR #149, opened 2026-02-04
Add desired_n parameter to power check API

mdulieu:
Extends the power check API endpoint to accept an optional `desired_n` parameter, enabling MDE (Minimum Detectable Effect) calculations from the frontend.

## Changes

- Add `desired_n` field to `BaseFrequentistDesignSpec` in `common_api_types.py`
- Pass `desired_n` to `check_power()` in `admin_api.py`

## Behavior

When `desired_n` is provided, the power check calculates the minimum detectable effect for that sample size instead of calculating the required sample size. When not provided, the original behavior is preserved (backward compatible).

Related to #148

### DIFF
--- a/src/xngin/apiserver/routers/admin/test_admin.py
+++ b/src/xngin/apiserver/routers/admin/test_admin.py
@@ -3472,7 +3472,7 @@ async def test_create_preassigned_experiment_with_table_name_and_primary_key(
 
     response = ppost(
         f"/v1/m/datasources/{ds_id}/experiments",
-        params={"chosen_n": 100, "random_state": 42},
+        params={"desired_n": 100, "random_state": 42},
         json={**request_json, "table_name": "dwh", "primary_key": "id"},
     )
     assert response.status_code == 200, response.content


poornimaramesh (commented):
the `create_experiment` endpoint in `admin_api.py` also takes the DesignSpec as input (inside `CreateExperimentRequest`), but also takes `chosen_n` as an input separately. 

Given that this is basically the same thing as `desired_n` do you think it would be a good idea to update the endpoint and all downstream functions using `chosen_n` also?
If you do decide to go down that route, it might blow up into a much bigger PR though -- so maybe we make this a ticket instead, and tackle it later?

mdulieu:
I think it makes sense to rename chosen_n to desired_n for consistency as they both mean the same thing. It looks like I would need to modify about 30-40 lines across ~10 files (mostly test updates). It is not a lot, I did it before and reverted actually. I'd like to include it in this PR if that's okay. What do you think?

> the `create_experiment` endpoint in `admin_api.py` also takes the DesignSpec as input (inside `CreateExperimentRequest`), but also takes `chosen_n` as an input separately.
> 
> Given that this is basically the same thing as `desired_n` do you think it would be a good idea to update the endpoint and all downstream functions using `chosen_n` also? If you do decide to go down that route, it might blow up into a much bigger PR though -- so maybe we make this a ticket instead, and tackle it later?

poornimaramesh:
> I think it makes sense to rename chosen_n to desired_n for consistency as they both mean the same thing. It looks like I would need to modify about 30-40 lines across ~10 files (mostly test updates). It is not a lot, I did it before and reverted actually. I'd like to include it in this PR if that's okay. What do you think?
> 
> > the `create_experiment` endpoint in `admin_api.py` also takes the DesignSpec as input (inside `CreateExperimentRequest`), but also takes `chosen_n` as an input separately.
> > Given that this is basically the same thing as `desired_n` do you think it would be a good idea to update the endpoint and all downstream functions using `chosen_n` also? If you do decide to go down that route, it might blow up into a much bigger PR though -- so maybe we make this a ticket instead, and tackle it later?

If you're ok making those changes, then I think you should go ahead! Happy to review again once you're done :)

poornimaramesh (approved):
Thanks so much @mdulieu -- this works!
I'm approving, but before merging, could you please raise a corresponding PR for the frontend?
Both `src/app/datasources/[datasourceId]/experiments/create/containers/frequent_ab/design-form.tsx` and `src/app/datasources/[datasourceId]/experiments/create/containers/mab/mab-metadata-form.tsx` both create payloads using `chosen_n`. If you could fix that and run `task genapiclient` (and push those changes to the frontend), this would be great.

Thanks again!

mdulieu:
> Thanks so much @mdulieu -- this works! I'm approving, but before merging, could you please raise a corresponding PR for the frontend? Both `src/app/datasources/[datasourceId]/experiments/create/containers/frequent_ab/design-form.tsx` and `src/app/datasources/[datasourceId]/experiments/create/containers/mab/mab-metadata-form.tsx` both create payloads using `chosen_n`. If you could fix that and run `task genapiclient` (and push those changes to the frontend), this would be great.
> 
> Thanks again!

@poornimaramesh I've created the corresponding frontend PR: https://github.com/agency-fund/evidential-fe/pull/220

The frontend changes include:
- Renamed `chosen_n` to `desired_n` in API parameters
- Renamed `chosenN` to `desiredN` in all TypeScript code
- Ran `task genapiclient` to regenerate the API client with the updated types
- All linting passed ✅

Both PRs should be ready to merge together!

PR #161, opened 2026-03-03
Integrate cluster analysis into power.py

mdulieu:
## Summary
Integrates cluster randomization support directly into the main power analysis functions in `power.py`.

## Changes
- Add optional cluster parameters (`icc`, `avg_cluster_size`, `cv`) to:
  - `_analyze_power_sample_size_mode()`
  - `calculate_mde_with_desired_n()`
- When cluster params provided, delegates to cluster functions
- When cluster params absent, uses existing individual randomization logic
- Fix cv parameter bug in `calculate_mde_cluster()`

## Testing
- 9 new integration tests with specific expected values
- All 39 tests pass
- Pre-commit checks pass

## Dependencies
- Built on top of #154 (cluster power analysis work)

## Related Issues
Part of #154 - Phase 2: Backend Power Calculations integration

PR #160, opened 2026-02-25
Add clustered test data generator with 1000 clusters

mdulieu:
## Summary
Implements next part of Phase 2 from #154: Generate realistic test data with 1000 clusters for comprehensive testing of cluster power analysis functions.

## Changes
- **New tool:** `generate_clustered_data.py` for creating realistic clustered datasets
  - Three clustering schemes: equal (`CV=0`), moderate (`CV ~ 0.3`), power-law (`CV~12`)
  - Four outcomes with specified ICCs: income, test_score, converted, engaged
  - Configurable via command-line arguments
  
- **Test dataset:** `clustered_dwh.csv.zst` (10K participants, 1K clusters)
  - Ready for testing cluster power functions
  - Compressed with zstd (264 KB)
  
- **Validation tests:** Comprehensive end-to-end validation
  - Verify generated ICCs match targets (0.04≈0.05, 0.19≈0.20, 0.21≈0.15)
  - Verify cluster CVs match design (0.00, 0.35, 11.87)

## Key Learnings
- Power-law clusters (extreme variation) still achieve good ICCs for continuous outcomes
- Binary outcomes with very low ICC (<0.05) are difficult to generate reliably

## Validation Results
```
income (equal clusters):     ICC=0.044 ✓ (target: 0.05)
test_score (powerlaw):       ICC=0.186 ✓ (target: 0.20)
engaged (powerlaw):          ICC=0.206 ✓ (target: 0.15)

Cluster CVs:
  Equal:     0.000 ✓
  Moderate:  0.347 ✓ (target: ~0.3)
  Power-law: 11.87 ✓ (target: >1.0)
```

## Testing
```bash
# Generate data
python tools/generate_clustered_data.py --n-participants 10000 --n-clusters 1000

# Run validation tests
pytest src/xngin/stats/test_cluster_icc.py::TestGeneratedClusteredData -v
```

PR #163, opened 2026-03-05
Add cluster randomization to power analysis API

mdulieu:
## Summary
Integrates cluster randomization into the power analysis API endpoint. Builds on the stats layer work to expose cluster randomization functionality through the API.

### API Layer (`src/xngin/apiserver/routers/admin/admin_api.py`)
- Updated `power_check()` endpoint to accept cluster parameters
- Two paths for cluster randomization:
  1. **Manual ICC**: User provides `icc`, `avg_cluster_size`, `cv`
  2. **Database ICC**: System calculates ICC from `cluster_column`
- Builds per-metric `cluster_params` dict before calling `check_power()`
- Uses `calculate_icc_and_cv()` when `cluster_column` is provided

### Request Model (`src/xngin/apiserver/routers/common_api_types.py`)
- **`PowerRequest`** - Added optional cluster parameter fields:
  - `cluster_column` - Column name for ICC calculation
  - `icc` - User-provided ICC value
  - `avg_cluster_size` - User-provided average cluster size  
  - `cv` - User-provided coefficient of variation (default 0.0)
  - Validation: `cluster_column` requires `table_name` and `primary_key`

### Response Model (`src/xngin/apiserver/routers/common_api_types.py`)
- **`PowerResponse`** - Updated type annotation:
  - Changed from: `list[MetricPowerAnalysis]`
  - To: `list[MetricPowerAnalysis | ClusterMetricPowerAnalysis]`
  - Enables serialization of cluster-specific fields in responses

### Test Infrastructure
- **`Taskfile.yml`** - Added `clustered_dwh` table to bootstrap task
  - Loads synthetic clustered test data (10K rows)
  - Required for database ICC calculation tests

### Integration Tests (`src/xngin/apiserver/routers/admin/test_admin.py`)
- **`test_power_check_with_manual_icc`** - Tests user-provided ICC path
  - Verifies cluster fields in response: `icc`, `avg_cluster_size`, `cv`, `design_effect`
  
- **`test_power_check_with_calculated_icc`** - Tests database-calculated ICC path
  - Verifies ICC calculated from `cluster_powerlaw` column
  - Validates calculated ICC is in expected range (~0.15-0.25)
  - Verifies high CV for power-law distribution (>2.0)

### Manual ICC (User provides values)
```python
POST /v1/m/datasources/{id}/power
{
  "design_spec": {...},
  "icc": 0.15,
  "avg_cluster_size": 30,
  "cv": 1.2
}
```

### Database ICC (System calculates from data)
```python
POST /v1/m/datasources/{id}/power
{
  "design_spec": {...},
  "table_name": "clustered_dwh",
  "primary_key": "participant_id",
  "cluster_column": "cluster_powerlaw"
}
```

## Response Format
```json
{
  "analyses": [{
    "metric_spec": {...},
    "target_n": 5640,
    "icc": 0.15,
    "avg_cluster_size": 30,
    "cv": 1.2,
    "design_effect": 5.35,
    "num_clusters_total": 188,
    "num_clusters_per_arm": 94
  }]
}
```

## Testing
- ✅ Manual ICC test passing
- ✅ Database ICC calculation test passing
- ✅ All 106 admin API tests passing
- ✅ Pre-commit checks passing
- ✅ Backward compatible (cluster params optional)

## Dependencies
Requires stats layer work from previous branches:
- `cluster_power.py` - Cluster power calculations
- `cluster_icc.py` - ICC/CV calculation from database
- Updates to `power.py` - Cluster params support


dctaf (commented):
Thanks! I've shared some comments below about project practices that might be helpful to this PR.

src/xngin/stats/cluster_icc.py
```diff
@@ -0,0 +1,128 @@
...
+
+    """
+    query = f"""
+        SELECT {cluster_column}, COUNT(*) as size
+        FROM {table_name}
+        WHERE {cluster_column} IS NOT NULL
+        GROUP BY {cluster_column}
+    """
+
+    df = pd.read_sql(query, session.bind)
```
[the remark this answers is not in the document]
mdulieu: Fixed - same approach as PR #159. Moved all SQL to queries.py using SQLAlchemy select() and .where() instead of f-strings.

src/xngin/stats/cluster_power.py
```diff
@@ -0,0 +1,229 @@
...
+        new_target_n = sum(n_per_arm_list)
+        effective_n = int(new_target_n / deff)
+
+        final_msg = individual_analysis.msg
+        if cv > 1.0 and final_msg:
+            warning = (
+                f"\n\nWarning: High cluster size variation (CV={cv:.2f}). Number of clusters estimates are approximate."
+            )
+
+            final_msg = MetricPowerAnalysisMessage(
```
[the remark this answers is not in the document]
mdulieu: Updated to use model_copy() in cluster_power.py:
```python
final_msg = final_msg.model_copy()
final_msg.msg += warning
final_msg.high_cluster_variation = True
```

src/xngin/stats/power.py
```diff
@@ -147,146 +170,163 @@ def _analyze_power_sample_size_mode(
...
-            ),
-        )
+    # Cluster analysis if parameters provided, otherwise individual
+    analysis: MetricPowerAnalysis | ClusterMetricPowerAnalysis
 
-    # Calculate target from pct_change if needed
-    if metric.metric_target is None and metric.metric_baseline is not None and metric.metric_pct_change is not None:
-        metric.metric_target = metric.metric_baseline * (1 + metric.metric_pct_change)
+    if icc is not None and avg_cluster_size is not None:
+        from xngin.stats.cluster_power import analyze_metric_power_cluster  # noqa: PLC0415
```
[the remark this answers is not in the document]
mdulieu: The inline import in power.py (line with # noqa: PLC0415) is intentional to avoid a circular import:
- power.py needs cluster_power.py for analyze_metric_power_cluster()
- cluster_power.py needs power.py for _analyze_power_sample_size_mode()

Moving it to the top would create a circular dependency. The noqa comment documents this is deliberate.

src/xngin/apiserver/routers/admin/admin_api.py
```diff
@@ -1907,6 +1943,45 @@ async def power_check(
...
+            for metric_stat in metric_stats:
+                cluster_stats = await asyncio.to_thread(
+                    calculate_icc_and_cv_from_database,
+                    dwh.session,
+                    sa_table,
+                    body.cluster_column,
+                    metric_stat.field_name,
+                    design_spec.filters,
+                )
+                cluster_params[metric_stat.field_name] = {
```
qixotic: Instead of passing this dict to check_power(), why not just set them in the `metric_stats` (list of DesignSpecMetric) that were fetched earlier via `get_stats_on_metrics`?

qixotic (changes requested):
After all the earlier PR merges (#155, #156, #159, #226 replacing #160, #161), looks like the relevant changes to this PR are: 
* e3bae7d47d36f23ade86f7a3aa4b49a7f510b335
* 1849a8ac618825e019bd9a0c192d9d04bb52a96d
* 6d26b9ad2b3e6f83dab4ca9131425bf2b345067f
  * Most of this commit was already merged in as part of earlier PRs, although fixes to admin_api.py look like they should be applied, as well as addition of `high_cluster_variation`.
* aebe35c3ced128534e9de65caa2a5deaeab31beb
* c222d06100098599f036a92ddf41d49bfc34ae5e

I think you could combine all the relevant additions and fixes still outstanding above into one new commit that combines it all, but applied to the latest head (using the updated interfaces) to replace the list here. If you have time in the next day or two to handle that @mdulieu let me know, else I'll finish it, thanks!

src/xngin/apiserver/routers/admin/admin_api.py
mdulieu: Replaced the previous history with a single commit on top of current main. The PR now contains only the still-relevant changes from your list (e3bae7d, 1849a8a, 6d26b9a's admin_api fixes + high_cluster_variation, aebe35c, c222d06), updated to use the current interfaces.

Also folded in your suggestion: cluster params (icc, avg_cluster_size, cv) are now set directly on the DesignSpecMetric objects in power_check, rather than building a separate cluster_params dict and passing it to check_power. The high_cluster_variation flag is set in cluster_power.py when cv > 1.0. Tests pass.

src/xngin/apiserver/routers/common_api_types.py
```diff
@@ -1087,6 +1088,44 @@ class PowerRequest(ApiBaseModel):
     ]
     primary_key: Annotated[str, Field(description="Primary key field name.")]
 
+    cluster_column: Annotated[
```
qixotic: This field is not just a property of a PowerRequest; it is actually a key field that should be part of the design spec, specifically the `BaseFrequentistDesignSpec`, as it's also required in the regressions for balance and analysis. It's presence should indicate to us that the experiment design is a cluster randomized one; not setting it means it randomizes at the the individual level.

We should also store the value in the db tables.py::Experiment table, so that your downstream change to update `analyze_experiment` can just pull the cluster_column from the the Experiment, but that can be a separate PR at this point.

src/xngin/apiserver/routers/common_api_types.py
```diff
@@ -1087,6 +1088,44 @@ class PowerRequest(ApiBaseModel):
...
+        str | None,
+        Field(
+            default=None,
+            description="Column name for cluster IDs. When provided, calculates ICC and CV from data "
+            "and performs cluster randomization power analysis.",
+        ),
+    ] = None
+
+    icc: Annotated[
```
qixotic: I think <icc, avg_cluster_size, cv> should be per-metric from the API point of view. That we may only allow users to set one tuple of values in the UI (for now?) should be separate from how we model this, else the model is not only overly restrictive but also potentially confusing to the reader.

Today's power logic relies on the metric names specified in the `BaseFrequentistDesignSpec.metrics`, which has the type `list[DesignSpecMetricRequest]`. The current design never let users specify data assumptions (metric_type, metric_baseline, metric_stddev), and instead only derived these from the data warehouse itself (as part of processing the `power_check`). The stats are combined with the info in the original request objects to produce a `DesignSpecMetric`, one per DesignSpecMetricRequest. (`DesignSpecMetricBase` holds fields common to DesignSpecMetricRequest and DesignSpecMetric. )

With this feature, we're proposing users be allowed to alternatively specify these data characteristics instead of deriving from the db. So these fields really belong in both the `DesignSpecMetricRequest` (if a user set their own), as well as `DesignSpecMetric` (copied over from DesignSpecMetricRequest or otherwise looked up from the dwh).  Therefore they could go in `DesignSpecMetricBase` as optional. I think you should be able to lift up the fields and validator you added to `DesignSpecMetric` in an earlier PR into the base class in that case.

src/xngin/apiserver/routers/admin/admin_api.py
```diff
@@ -1907,6 +1908,34 @@ async def power_check(
             design_spec.filters,
         )
 
+        # Handle cluster randomization parameters
```
qixotic: All of this extra logic really should be conditional on a `cluster_column` param being set on the design spec, as I noticed below in src/xngin/apiserver/routers/common_api_types.py.

src/xngin/apiserver/routers/admin/admin_api.py
```diff
@@ -1907,6 +1908,34 @@ async def power_check(
...
+        if body.icc is not None and body.avg_cluster_size is not None:
+            # User-supplied cluster params: same values for all metrics
```
qixotic: It seems to be a strong assumption to treat ICC as the same for all metrics supplied. I think it would be better if we allow for the flexibility of having per-metric values for ICC in the model (which is possible by lifting all the fields into the DesignSpecMetricBase), but in the _frontend_ specifically use the same values for all metrics when making the design spec (and its list of DesignSpecMetricRequests) while telling the user that, if we only wish to support the primary metric's ICC for now.

src/xngin/apiserver/routers/admin/admin_api.py
```diff
@@ -1907,6 +1908,34 @@ async def power_check(
...
+        elif body.cluster_column is not None:
+            # Compute per-metric ICC/CV from the database
+            if not design_spec.metrics:
+                raise HTTPException(
+                    status_code=status.HTTP_400_BAD_REQUEST,
+                    detail="At least one metric required for cluster analysis",
+                )
+            for metric_stat in metric_stats:
+                cluster_stats = await asyncio.to_thread(
+                    calculate_icc_and_cv_from_database,
```
qixotic: A side note: this function calls `get_cluster_size_stats`, but those stats including the avg_cluster_size and cv are independent of the metrics being tracked, so it's inefficient to repeatedly call this for each metric. In the future, we could optimize `calculate_icc_from_dataframe` to handle ICC for multiple metrics separately from the cluster stats. And since we're currently using ANOVA to compute the ICC, I think we could even push the component aggregates into the db and avoid getting the raw outcome data, and do that for all metrics in one query analogous to how `get_stats_on_metrics` handles multiple metrics.  (That would only be necessary then if we went back to using MixedLM. re: https://github.com/agency-fund/evidential-be/pull/159/changes/fc431ec4ac55d1ef52161e8f0f49de439b0eab81)

But please don't worry about any of these optimizations in this PR! We can do that in a followup; let's get the basics working first.

qixotic (changes requested):
Hi @mdulieu , I pushed one change to your remote branch in 6133c2df330dc4c173143bf790764c0e04d5eb72 that uses the new power_adapters.py module that holds `calculate_icc_and_cv_from_database` now (originally from #161). Please sync those changes to your local branch before addressing these comments. Thanks!

src/xngin/apiserver/routers/common_api_types.py
qixotic: I will take care of the table.py change in a future PR right after this PR gets merged, so just focus on all the API model-level changes in this PR, thanks!

mdulieu: Moved cluster_column onto BaseFrequentistDesignSpec. Left the Experiment table change for your follow-up PR.

src/xngin/apiserver/routers/common_api_types.py
mdulieu: Lifted icc, avg_cluster_size, cv (and the all-or-none validator) into DesignSpecMetricBase. Both DesignSpecMetricRequest and DesignSpecMetric now inherit them, so callers can supply per-metric values directly.

src/xngin/apiserver/routers/admin/admin_api.py
mdulieu: The whole cluster block is now gated on design_spec.cluster_column. If unset, no cluster logic runs.

src/xngin/apiserver/routers/admin/admin_api.py
mdulieu: power_check now reads (icc, avg_cluster_size, cv) per metric from each DesignSpecMetricRequest, falling back to calculate_icc_and_cv_from_database per metric when not supplied.

src/xngin/apiserver/routers/admin/admin_api.py
mdulieu: ok, leaving as is. Happy to do the optimization as a follow-up.

qixotic:
It looks like you had a bad merge. I've fixed it, but please double check that you have installed the pre-commit command with a `uv sync` and that the precommit hooks into your local git with: `uv run pre-commit install`

I've also renamed cluster_column => cluster_key, and fixed some indenting.

This should be mergable now; will do so later in the day, thanks!

PR #201, opened 2026-04-01
feat: support user-supplied desired_n in power check

mdulieu:
## Summary

Extends the power check endpoint to support a user-supplied `desired_n`, returning the achievable MDE alongside the normal power check results. This is in support of issue https://github.com/agency-fund/evidential-be/issues/216.

## Motivation

Previously, passing `desired_n` to the power check switched the calculation into MDE-only mode, discarding `target_n` and `sufficient_n`. This made it impossible for the frontend to show both the normal power check result and the estimated MDE for a user-chosen sample size at the same time.

## Changes

**`common_api_types.py`**
- ~Added `desired_n: int | None` to `PowerRequest` so the frontend can pass a user-chosen sample size~
- Added `pct_change_with_desired_n: float | None` to `MetricPowerAnalysis` to hold the resulting MDE

**`power.py`**
- Updated `check_power` to always run the normal sample size calculation first, then run the MDE calculation as a second step if `desired_n` was provided
- Result: `target_n`, `sufficient_n`, and `pct_change_with_desired_n` are all present on the same output object

**`admin_api.py`**
- ~Updated the power check endpoint to pass `body.desired_n` instead of `design_spec.desired_n`~
- Removed query param `desired_n` from `create_experiment` and regenerated **`admin_api_client.py‎`**

**`tables.py`**
- persists `desired_n` in Experiments

Updated various tests to reflect the changed behaviors.

## Behavior

- When `desired_n` is not provided: behavior is unchanged
- When `desired_n` is provided: normal power check runs as usual, and `pct_change_with_desired_n` is additionally populated in a best-effort manner with the MDE achievable at that sample size (regardless of real availability). If it encounters an internal error e.g. due to a data validation issue, `pct_change_with_desired_n` is left `None`.

## Follow-up

Frontend PR will wire up `desired_n` in the power check page and display `pct_change_with_desired_n` as the estimated MDE. Cluster support to be added in a separate PR. 

Extend PowerRequest to support reuse of metric stats in DesignSpecMetric. 

Drop `n_trials` from Experiments table (unused by bandits).


src/xngin/stats/power.py
```diff
@@ -425,7 +425,18 @@ def check_power(
     analyses = []
     for metric in metrics:
         try:
-            analyses.append(analyze_metric_power(metric, n_arms, power, alpha, arm_weights, desired_n))
+            analysis = analyze_metric_power(metric, n_arms, power, alpha, arm_weights)
+            if desired_n is not None:
+                _, pct_change = calculate_mde_with_desired_n(
```
qixotic: I've modified this as part of a rebase to just call `analyze_metric_power` again but with the desired_n passed in. Same idea as what you're doing, but this also leverages some data validation that analyze_metric_power does when in MDE mode.

qixotic (commented):
Updated this PR to always use desired_n from the DesignSpec, and removed the API's query param on creation of experiments. Also added persistence to our Experiments table of this value and backfill it for preassigned experiments. @dctaf could you take a look?

Corresponding FE updates are in https://github.com/agency-fund/evidential-fe/pull/244

qixotic:
Also confirmed with @poornimaramesh that we don't need c67a845882e58deac7103f34099cd0e71120ac2e
> we can toss n_trials, we don't use it for the bandits!

src/xngin/stats/power.py
```diff
@@ -125,21 +125,30 @@ def check_power(
...
                     metric=metric,
                     n_arms=n_arms,
                     arm_weights=arm_weights,
                     desired_n=desired_n,
                     power=power,
-                    alpha=alpha,
-                )
-            )
+                    alpha=alpha)
+                analysis.pct_change_with_desired_n = mde_analysis.pct_change_possible
```
dctaf: 5.4-medium flagged an issue with this but I won't share the specifics because it was unintelligible. But here's what I think it is:

We don't have detailed documentation on how to interpret these values on the types, and they could be improved, and I'm possibly confused about what we want to return in failure cases, but it seems to me that we would want `.pct_change_with_desired_n` to be updated at the same time as other values such as `.msg`, `.msg.type`, and `.pct_change_possible`.

When the first pass analysis fails -- do we still want to return all of the other fields as set by analyze_power_metric?

Here's an illustrative DesignSpec which will fail the first analysis pass:

```python
    design_spec = PreassignedFrequentistExperimentSpec(
        experiment_type=ExperimentsType.FREQ_PREASSIGNED,
        experiment_name="test power check desired_n failure",
        description="design_spec.desired_n should surface MDE validation errors",
        table_name="dwh",
        primary_key="id",
        start_date=datetime(2024, 1, 1, tzinfo=UTC),
        end_date=datetime.now(UTC) + timedelta(days=1),
        arms=[
            Arm(arm_name="control", arm_description="Control group"),
            Arm(arm_name="treatment", arm_description="Treatment group"),
        ],
        metrics=[DesignSpecMetricRequest(field_name="is_onboarded_onetime", metric_pct_change=0.1)],
        strata=[],
        filters=[
            Filter(
                field_name="sample_timestamp",
                relation=Relation.BETWEEN,
                value=["2024-12-05T19:19:28Z", "2024-12-05T19:19:28Z"],
            ),
        ],
        desired_n=500,
    )
```

I recommend putting that into a unit test and confirming the output.

src/xngin/apiserver/routers/admin/test_admin_api.py
```diff
@@ -1618,6 +1616,44 @@ async def test_power_check_with_unbalanced_arms(testing_datasource, aclient: Adm
...
+    assert power_response.analyses[0].target_n == 474  # min sample size
+
+    # And when not set, we get only the default minimum sample size calculation.
+    design_spec_plain = design_spec.model_copy(update={"desired_n": None})
+    power_response_plain = aclient.power_check(
+        datasource_id=testing_datasource.ds.id,
+        body=PowerRequest(design_spec=design_spec_plain),
+    ).data
+    assert power_response_plain.analyses[0].pct_change_with_desired_n is None
+    assert power_response_plain.analyses[0].target_n == 474
```
dctaf: The MetricPowerAnalysisMessage type has a `type` field which we should probably check here too. This is what the UI uses to determine success and failure, and it would be good to pin the behavior of that field.

src/xngin/stats/power.py
```diff
@@ -125,21 +125,23 @@ def check_power(
             instead of the minimum sample size for a desired effect. Applies to all metrics.
 
     Returns:
-        List of MetricPowerAnalysis results
+        List of MetricPowerAnalysis results, always including the minimum sample size for a desired
+        effect, and optionally including the MDE for the desired_n if provided.
```
dctaf: This comment could be reworded.

1. Can we refer to field names specifically instead of "minimum sample size", e.g. "target_n will be populated in all items."
2. Consider clarifying that it is one item per metric, in the same order as `metrics`.

dctaf (changes requested):

src/xngin/stats/power.py
qixotic: Thanks, I've tightened the logic around the 2nd call to analyze_metric_power() to ensure behavior of the desired_n's MDE calc is best-effort, so that it is provided if the first call still completes (even if there are insufficient available samples, as the MDE calc always assumes possible sufficiency). This is supported by more unittests.

src/xngin/stats/power.py
```diff
@@ -125,21 +125,45 @@ def check_power(
...
+                        arm_weights=arm_weights,
+                        desired_n=desired_n,
+                        power=power,
+                        alpha=alpha,
+                    )
+                    # pct_change_possible may be None if there was any error created with
+                    # power_analysis_error().
+                    if mde_analysis.pct_change_possible is not None:
+                        analysis.pct_change_with_desired_n = mde_analysis.pct_change_possible
+                except ValueError:
```
dctaf, line 161: ValueError is a wide exception to catch. Is it interesting enough to the caller that the optional MDE cannot be computed that we should report it to them?

dctaf (approved):

src/xngin/stats/power.py
qixotic: Yeah I thought about that, but don't want to complicate the response any further for now. Additionally, the checks in the solve-for-MDE case for metric_baseline & metric_type & metric_stddev are also done in `solve_for_sample_size_individual` first so would be redundant, and for the clustered case (icc & avg_cluster_size), we really rely on pydantic model validation. So the ValueErrors we raise internally are all sanity checks. The frontend should treat a response of None as an error that is generically reported back, e.g. "Unable to compute MDE".

(Note that in theory one might want to run an experiment on a bunch of samples that are all null-valued, in which case both should give a power check error message, but as long as the requested amount is <= available_n, we should let it through under the assumption the user knows what they're doing as we do today.)

I would also like to refactor things further in the future to try and clean up some of the error handling for readability between the indiv and clustered paths.

Thanks for the review!

PR #204, opened 2026-04-07
Restrict CORS allowed origins.

dctaf:
The CORS policy previously used allow_origins=["*"], permitting any website
to make cross-origin requests to the API with arbitrary headers including
Authorization.

Behavior by environment:
- Dev (ENVIRONMENT unset or "dev"): defaults to ["*"] as before, so local
  development workflows are unaffected.
- Non-dev: CORS_ALLOWED_ORIGINS must be set explicitly. If it is missing or
  empty the server refuses to start, preventing accidental wildcard exposure
  in production due to a misconfigured deployment.


qixotic (approved):
Thanks for setting the var already in prod

PR #209, opened 2026-04-09
Migrate tq to template strings and add some basic tq tests.

dctaf:
- **Use psycopg template strings in task queue**
- **Add some tests for task queue.**

### DIFF
--- a/src/xngin/tq/test_tq.py
+++ b/src/xngin/tq/test_tq.py
@@ -4,7 +4,7 @@ import threading
 import time
 from collections.abc import Callable, Generator
 from contextlib import contextmanager
-from datetime import UTC, datetime
+from datetime import timedelta
 
 import psycopg
 import pytest
@@ -153,7 +153,6 @@ async def test_task_queue_requeues_failed_task_with_backoff(
 
     task_queue.register_handler("test.retry", handler)
     with tq_runner(task_queue):
-        inserted_at = datetime.now(UTC)
         task = await insert_task(
             xngin_session,
             task_type="test.retry",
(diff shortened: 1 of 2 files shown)


src/xngin/tq/test_tq.py
```diff
@@ -0,0 +1,200 @@
...
+            task.id,
+            "pending",
+            predicate=lambda row: row.retry_count == 1 and row.message == "handler failed",
+        )
+
+        assert handled_task_ids.get(timeout=1) == task.id
+        assert handled_task_ids.empty()
+        assert pending_task.retry_count == 1
+        assert pending_task.message == "handler failed"
+        assert pending_task.embargo_until > inserted_at
```
qixotic: could check that it's at least 15 min later as currently fixed in https://github.com/agency-fund/evidential-be/blob/main/src/xngin/tq/task_queue.py#L186 yes?

qixotic (approved):

src/xngin/tq/test_tq.py
```diff
@@ -0,0 +1,200 @@
...
+from xngin.tq import task_queue as task_queue_module
+from xngin.tq.task_queue import Task, TaskQueue
+
+pytest_plugins = ("xngin.apiserver.conftest",)
+
+STATUS_TIMEOUT_SECS = 5.0
+
+
+@dataclass(slots=True)
+class TqHandle:
```
qixotic: not used?

src/xngin/tq/test_tq.py
dctaf: I've strengthened this test. The first retry is at 1 minute.

src/xngin/tq/test_tq.py
dctaf: Removed. Thanks!

PR #210, opened 2026-04-09
fix: validate webhook hostnames when creating/updating/using them

qixotic:
Prevents possible SSRF risk by having admin_api_types.py::validate_webhook_url() also check if the hostname resolves to a safe IP address, and similarly do that check every time we use the webhook to deliver events. Updates existing tests accordingly and adds 2 more to verify we're raising the expected LateValidationError when the hostname fails the safety check.

### DIFF
--- a/src/xngin/apiserver/routers/admin/admin_api_types.py
+++ b/src/xngin/apiserver/routers/admin/admin_api_types.py
@@ -7,7 +7,9 @@ from annotated_types import Ge, Le
 from pydantic import BaseModel, ConfigDict, Field, field_validator
 
 from xngin.apiserver.common_field_types import FieldName
+from xngin.apiserver.dns.safe_resolve import DnsLookupError, safe_resolve
 from xngin.apiserver.dwh.inspection_types import FieldDescriptor, ParticipantsSchema
+from xngin.apiserver.exceptions_common import LateValidationError
 from xngin.apiserver.limits import (
     MAX_LENGTH_OF_DESCRIPTION_VALUE,
     MAX_LENGTH_OF_EMAIL_VALUE,
@@ -36,9 +38,13 @@ def validate_webhook_url(url: str) -> str:
     """Validates that a URL is a properly formatted HTTP or HTTPS URL."""
     parsed = urlparse(url)
     if not parsed.scheme or parsed.scheme not in {"http", "https"}:
-        raise ValueError("URL must use http or https scheme")
+        raise LateValidationError("URL must use http or https scheme")
     if not parsed.netloc:
-        raise ValueError("URL must include a valid domain")
+        raise LateValidationError("URL must include a valid domain")
+    try:
+        safe_resolve(parsed.hostname)
+    except DnsLookupError as e:
+        raise LateValidationError(f"URL hostname failed safety validation: {parsed.hostname}") from e
     return url
 
 
--- a/src/xngin/apiserver/routers/admin/admin_api_types.py
+++ b/src/xngin/apiserver/routers/admin/admin_api_types.py
@@ -9,7 +9,6 @@ from pydantic import BaseModel, ConfigDict, Field, field_validator
 from xngin.apiserver.common_field_types import FieldName
 from xngin.apiserver.dns.safe_resolve import DnsLookupError, safe_resolve
 from xngin.apiserver.dwh.inspection_types import FieldDescriptor, ParticipantsSchema
-from xngin.apiserver.exceptions_common import LateValidationError
 from xngin.apiserver.limits import (
     MAX_LENGTH_OF_DESCRIPTION_VALUE,
     MAX_LENGTH_OF_EMAIL_VALUE,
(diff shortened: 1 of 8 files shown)


src/xngin/apiserver/routers/admin/admin_api_types.py
```diff
@@ -36,9 +38,13 @@ def validate_webhook_url(url: str) -> str:
     """Validates that a URL is a properly formatted HTTP or HTTPS URL."""
     parsed = urlparse(url)
     if not parsed.scheme or parsed.scheme not in {"http", "https"}:
-        raise ValueError("URL must use http or https scheme")
+        raise LateValidationError("URL must use http or https scheme")
```
dctaf: Custom Pydantic validators should constrain the set of exceptions they raise to ValueError, AssertionError, or PydanticCustomError:

https://pydantic.dev/docs/validation/latest/concepts/validators/#raising-validation-errors

src/xngin/apiserver/routers/admin/admin_api_types.py
```diff
@@ -36,9 +38,13 @@ def validate_webhook_url(url: str) -> str:
...
     if not parsed.netloc:
-        raise ValueError("URL must include a valid domain")
+        raise LateValidationError("URL must include a valid domain")
+    try:
+        safe_resolve(parsed.hostname)
+    except DnsLookupError as e:
+        raise LateValidationError(f"URL hostname failed safety validation: {parsed.hostname}") from e
```
dctaf: DnsLookupError can be raised for a variety of reasons. I suggest using a more general description, such as "Failed to resolve hostname from webhook URL."

src/xngin/tq/handlers.py
```diff
@@ -34,6 +37,13 @@ def webhook_outbound_handler(task: Task):
         request = WebhookOutboundTask.model_validate(task.payload)
         logger.info(f"Processing {request}")
 
+        parsed_url = urlparse(request.url)
+        try:
+            safe_resolve(parsed_url.hostname)
+        except DnsLookupError as e:
+            logger.warning(f"Webhook URL failed safety check at delivery time: {request.url} ({e})")
+            raise
```
dctaf: 1. The "Events" UI allows users to see the results of outbound webhooks. It describes successes and failures. The failures are captured in the events table via the `except httpx.HttpError` block a few lines down. We should validate the hostname within that try/except block (or add another one) because it will allow us to log the DNS resolution failure and make that visible to the user in the UI.
2. httpx will make its own DNS request and may see different results than the lookup from safe_resolve. The potential for difference between the two is what enables DNS rebinding attacks. The way to solve that is to tell the HTTP client to use the validated IP address from the safe_resolve() call. The HTTPX documentation describes one way to do this (https://www.python-httpx.org/advanced/extensions/#sni_hostname). Until we replace httpx, that's probably the way to go.

src/xngin/apiserver/routers/admin/admin_api_types.py
qixotic: thanks, done

src/xngin/tq/handlers.py
qixotic: done, thanks

src/xngin/tq/handlers.py
```diff
@@ -34,49 +57,73 @@ def webhook_outbound_handler(task: Task):
...
-                    timeout=10.0,
-                    headers=request.headers,
-                )
+                scheme = parsed.scheme
+                hostname = parsed.hostname
+                safe_ip = safe_resolve(hostname)
+                assert hostname is not None  # safe_resolve() checks this but mypy doesn't know that
+                port = parsed.port
+
+                ip_host = f"[{safe_ip}]" if ":" in safe_ip else safe_ip
```
dctaf: Note: This implements only partial support for IPv6. safe_resolve() only looks up A records (IPv4), not AAAA records (IPv6), so we will never hit the `f"[...]"` case. Consider removing for clarity.

src/xngin/tq/handlers.py
```diff
@@ -34,49 +57,73 @@ def webhook_outbound_handler(task: Task):
...
+                ip_host_port = f"{ip_host}:{port}" if port else ip_host
+                connect_url = urlunparse((
```
dctaf, line 74: If the user provided a webhook with credentials (e.g. `https://infra:password@hostname/...`), this reconstruction will remove them. I am not aware of any reason we should disallow that, but I also don't know if httpx will use them anyway. If it is trivial to retain these, I recommend we do!

src/xngin/tq/test_handlers.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+async def test_webhook_outbound_handler_records_dns_failure_event(
+    xngin_session: AsyncSession,
+    tq_dsn: str,
+):
+    org = tables.Organization(name="test-org")
+    xngin_session.add(org)
+    await xngin_session.commit()
+    await xngin_session.refresh(org)
```
dctaf: The tests in this file can use `aclient: AdminAPIClient` AFAIK and I recommend we prefer doing so any time we deal with tables other than Task. Organizations, Events, and Webhooks can rely on the full infrastructure.

test_tq.py does not use AdminAPIClient because it has very narrow dependencies -- it tests only the tq event loop, so it only depends on the things that the event loop does:
- the `tasks` table existing
- all database operations use SQL directly (not SQLAlchemy)
- none of it depends on the xngin server implementation or business logic

dctaf (approved):
Thorough tests, nice! Please address the comments below at your discretion and merge at will!

PR #211, opened 2026-04-09
fix: prevent possible sql injection via search_path and table_name

qixotic:
Properly quote the user-supplied table_name using sqlalchemy's table() construct.
Add extra pydantic model constraints on search_path, and construct the query with the SQLA dialect's IdentifierPreparer.

### DIFF
--- a/src/xngin/apiserver/dwh/dwh_session.py
+++ b/src/xngin/apiserver/dwh/dwh_session.py
@@ -147,15 +147,27 @@ class DwhSession:
             raise RuntimeError("DwhSession not entered - use 'async with DwhSession(...) as dwh:'")
         return self._engine
 
-    def _inspect_table_blocking(self, table_name: str, use_sa_autoload: bool | None = None) -> sqlalchemy.Table:
+    def _inspect_table_blocking(
+        self,
+        table_name: str,
+        *,
+        schema_name: str | None = None,
+        use_sa_autoload: bool | None = None,
+    ) -> sqlalchemy.Table:
         if use_sa_autoload is None:
             use_sa_autoload = self.dwh_config.supports_sa_autoload()
         metadata = sqlalchemy.MetaData()
         try:
             if use_sa_autoload:
-                return sqlalchemy.Table(table_name, metadata, autoload_with=self._safe_engine(), quote=False)
+                return sqlalchemy.Table(
+                    table_name,
+                    metadata,
+                    schema=schema_name,
+                    autoload_with=self._safe_engine(),
+                    quote=False,
+                )
             # This method of introspection should only be used if the db dialect doesn't support Sqlalchemy2 reflection.
-            return self._inspect_table_from_cursor_blocking(self._safe_engine(), table_name)
+            return self._inspect_table_from_cursor_blocking(self._safe_engine(), table_name, schema_name)
         except sqlalchemy.exc.ProgrammingError:
             logger.exception("Failed to create a Table! use_sa_autoload: {}", use_sa_autoload)
             raise
@@ -165,7 +177,7 @@ class DwhSession:
             raise CannotFindTableError(table_name, existing_tables) from nste
 
     def _inspect_table_from_cursor_blocking(
-        self, engine: sqlalchemy.engine.Engine, table_name: str
+        self, engine: sqlalchemy.engine.Engine, table_name: str, schema_name: str | None = None
     ) -> sqlalchemy.Table:
         """Creates a SQLAlchemy Table instance from cursor description metadata."""
 
@@ -173,7 +185,7 @@ class DwhSession:
         metadata = sqlalchemy.MetaData()
         try:
             with engine.begin() as connection:
-                query = query_constructors.create_inspect_table_from_cursor_query(table_name)
+                query = query_constructors.create_inspect_table_from_cursor_query(table_name, schema_name)
                 result = connection.execute(query)
                 description = result.cursor.description
                 for col in description:
@@ -233,18 +245,23 @@ class DwhSession:
         .name, .type, and .nullable.
 
         Args:
-            table_name: Name of the table to inspect
+            table_name: Name of the table to inspect. Only unqualified table names are supported.
             use_sa_autoload: Whether to use SQLAlchemy reflection. If None, uses config default.
 
         Returns:
             SQLAlchemy Table object
         """
-        return await asyncio.to_thread(self._inspect_table_blocking, table_name, use_sa_autoload)
+        return await asyncio.to_thread(
+            self._inspect_table_blocking,
+            table_name,
+            schema_name=None,
+            use_sa_autoload=use_sa_autoload,
+        )
 
     def _inspect_table_with_descriptors_blocking(
         self, table_name: str, unique_id_field: str, use_sa_autoload: bool | None = None
     ) -> InspectTableWithDescriptorsResult:
-        sa_table = self._inspect_table_blocking(table_name, use_sa_autoload)
+        sa_table = self._inspect_table_blocking(table_name, schema_name=None, use_sa_autoload=use_sa_autoload)
         db_schema = generate_field_descriptors(sa_table, unique_id_field)
         return InspectTableWithDescriptorsResult(sa_table=sa_table, db_schema=db_schema)
 
@@ -288,7 +305,7 @@ class DwhSession:
         n: int,
         use_sa_autoload: bool | None = None,
     ) -> GetParticipantsResult:
-        sa_table = self._inspect_table_blocking(table_name, use_sa_autoload)
+        sa_table = self._inspect_table_blocking(table_name, schema_name=None, use_sa_autoload=use_sa_autoload)
         participants = self._query_for_participants_blocking(sa_table, select_columns, filters, n)
         return GetParticipantsResult(sa_table=sa_table, participants=participants)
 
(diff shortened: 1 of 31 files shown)


src/xngin/apiserver/dwh/dwh_session.py
```diff
@@ -420,7 +418,8 @@ def set_search_path(dbapi_connection, _connection_record):
                 existing_autocommit = dbapi_connection.autocommit
                 dbapi_connection.autocommit = True
                 cursor = dbapi_connection.cursor()
-                cursor.execute(f"SET SESSION search_path={self.dwh_config.search_path}")  # type: ignore[union-attr]
+                path = getattr(self.dwh_config, "search_path", "public")
```
dctaf: The use of getattr() might reduce clarity: the if-condition on 414 ensures .search_path is truthy, and the presence of "default" misleads the reader into thinking it would be used. Can we use `self.dwh_config.search_path` directly?

src/xngin/apiserver/settings.py
```diff
@@ -258,6 +258,11 @@ def to_sqlalchemy_url(self) -> sqlalchemy.URL:
         )
 
 
+# Allows comma-separated PostgreSQL schema identifiers (alphanumeric, _ and $).
+# Used in both Dsn and the admin API request models to prevent SQL injection.
+SEARCH_PATH_PATTERN = r"^[a-zA-Z_][a-zA-Z0-9_$]*(,\s*[a-zA-Z_][a-zA-Z0-9_$]*)*$"
```
dctaf: There are demons here and we need to decide whether to parse the string properly or to constrain to a subset of valid search path strings. `$user` is a valid schema and is rejected by this pattern. Also, quoted Postgres identifiers can contain commas, so the split-on-commas approach isn't technically correct and we would need to use a real parser to validate search paths using identifiers that require quotes.

src/xngin/apiserver/dwh/query_constructors.py
```diff
@@ -125,6 +126,28 @@ def create_filter(col: sqlalchemy.Column, filter_: Filter) -> ColumnElement:
             raise RuntimeError("Bug: invalid Filter.")
 
 
+def create_inspect_table_from_cursor_query(table_name: str) -> sqlalchemy.Select:
+    """Returns a zero-row SELECT used to read column metadata via cursor.description.
+
+    We use SQLAlchemy's identifier quoting to prevent table-name-based SQL injection
+    (e.g. "foo; DROP TABLE bar" becomes a quoted identifier rather than executable SQL).
+    """
+    return select(literal_column("*")).select_from(sa_table(table_name)).limit(0)
```
dctaf: This will break if table_name is qualified with a schema name, e.g.  `"other_schema".my_table`. Our UI requires users to select the unqualified table name from a dropdown, but it might be useful for future reference to note that this method only works with unqualified table names.

dctaf (changes requested):

src/xngin/apiserver/dwh/query_constructors.py
qixotic: Modified the function to explicitly take a schema, although note that we aren't using that right now.

src/xngin/apiserver/settings.py
qixotic: Yeah, ok; I've reverted this bit and instead we'll use pg's [set_config](https://www.postgresql.org/docs/current/functions-admin.html#FUNCTIONS-ADMIN-SET) directly. When we have to support other db types' similar concepts, we can move the query+params into query_constructors.py and switch off the dialect for the backend-specific approaches.

src/xngin/apiserver/dwh/dwh_session.py
```diff
@@ -414,13 +430,20 @@ def _extra_engine_setup(self, engine: Engine):
         """
         # Handle search_path for PostgreSQL
         if isinstance(self.dwh_config, Dsn) and self.dwh_config.search_path:
+            search_path_sql_arg = self.dwh_config.search_path
+            assert search_path_sql_arg is not None  # type-narrowing for mypy
```
dctaf: My local `uv run mypy src` does not require this assertion. Does yours? 

```
$ uv run mypy --version
mypy 1.19.1 (compiled: yes)
```

src/xngin/apiserver/dwh/query_constructors.py
```diff
@@ -96,6 +96,16 @@ def create_filter(col: sqlalchemy.Column, filter_: Filter) -> ColumnElement:
...
 
 
+def create_inspect_table_from_cursor_query(table_name: str, schema_name: str | None = None) -> sqlalchemy.Select:
+    """Returns a zero-row SELECT used to read column metadata via cursor.description (driver-dependent).
+
+    We use SQLAlchemy's identifier quoting to prevent table-name-based SQL injection
+    (e.g. "foo; DROP TABLE bar" becomes a quoted identifier rather than executable SQL).
+    """
+    table = Table(table_name, sqlalchemy.MetaData(), schema=schema_name)
+    return select(table).limit(0)
```
dctaf: We are no longer generating the `select(text("*"))` workaround we were using for Redshift's lack of sa_autoload support. Is that not required anymore?

src/xngin/apiserver/dwh/dwh_session.py
```diff
@@ -147,15 +148,27 @@ def _safe_engine(self) -> Engine:
             raise RuntimeError("DwhSession not entered - use 'async with DwhSession(...) as dwh:'")
         return self._engine
 
-    def _inspect_table_blocking(self, table_name: str, use_sa_autoload: bool | None = None) -> sqlalchemy.Table:
+    def _inspect_table_blocking(
+        self,
+        table_name: str,
+        *,
+        schema_name: str | None = None,
```
dctaf: Now that we are always reading schema name from the config object, and that all callers are passing schema_name=None, I think readability would improve if we remove the schema_name argument from the dwh_session methods entirely.  

Also, because DwhSession accepts a dwh_config as a constructor argument, and we are not doing any pre-processing on it, it will always be available to methods on that class in self.dwh_config.

dctaf (changes requested):

src/xngin/apiserver/settings.py
dctaf: Oh, nice. `set_config` makes sense here!

src/xngin/apiserver/dwh/dwh_session.py
qixotic: Ah yes, capturing the search path outside the listener guarantees the type, whereas referencing the search_path via the dwh_config is not due to the deferred execution, and presumably self.dwh_config could change from under us by the time the listener is called.

src/xngin/apiserver/dwh/query_constructors.py
qixotic: whoops still need that. select(table) works only if it already knows of the Columns to expand out

src/xngin/apiserver/dwh/dwh_session.py
qixotic: done

src/xngin/apiserver/dwh/query_constructors.py
```diff
@@ -126,26 +96,13 @@ def create_filter(col: sqlalchemy.Column, filter_: Filter) -> ColumnElement:
...
-    """
-    return select(literal_column("*")).select_from(sa_table(table_name)).limit(0)
-
-
-def build_search_path_sql(preparer: sqlalchemy.sql.compiler.IdentifierPreparer, raw_search_path: str) -> str:
-    """Builds a SET SESSION search_path=... statement with quoted identifiers.
-
-    Uses the provided SQLA engine dialect's IdentifierPreparer to handle double-quote escaping and
-    identifier length limits.
+def create_inspect_table_from_cursor_query(table_name: str, schema_name: str | None = None) -> sqlalchemy.Select:
```
dctaf, line 99: The only place we are passing a non-none schema_name is from a test. What is the expected behavior for Redshift warehouses with a search path?

dctaf, line 99: Consider inlining this method. It's only two lines and it might make reading the implementation of its only caller (`_inspect_table_from_cursor_blocking`) easier.

dctaf (changes requested):

src/xngin/apiserver/dwh/query_constructors.py
qixotic: I'd like to keep it to avoid future questions about "what if table_name is qualified with a schema name?". And its' a good reminder that Table(..., schema=) is the most backend agnostic way to do this in SQLA if we ever have to.

qixotic: Redshift still supports the set_config() function from pg; I've confirmed as part of #237 merged into this that we'll use the search_path correctly.

dctaf (approved):

src/xngin/apiserver/dwh/dwh_session.py
```diff
@@ -331,9 +344,11 @@ def _list_tables_blocking(self) -> list[str]:
             if isinstance(self.dwh_config, Dsn) and self.dwh_config.is_redshift():
                 query = text(
                     "SELECT table_name FROM information_schema.tables "
-                    "WHERE table_schema IN (:search_path) ORDER BY table_name"
+                    "WHERE table_schema = ANY(current_schemas(false)) "
+                    "AND table_type IN ('BASE TABLE', 'VIEW') "
+                    "ORDER BY table_name"
```
qixotic, line 349: @dctaf mentioned this won't work with s3 tables, and gpt5.5 noted the same earlier in a caveat:
> depending on whether you also want late-binding views or external views, information_schema.tables may not cover every Redshift view-like object. For normal user-created views, table_type = 'VIEW' is the simple fix.

Indeed, there are table types that we won't be able to support this way, but should be left to a future ticket if we need that support (hope we do one day!). Specifically, the following more-Redshift-specific query should work to pick up external schema/tables from e.g. Redshift Spectrum S3 tables or datashares (cross-cluster/account sharing of tables):

```
SELECT
    schema_name,
    table_name,
    table_type
FROM SVV_ALL_TABLES
WHERE
  schema_name = ANY(current_schemas(false)) AND
  table_type IN ('TABLE', 'VIEW', 'EXTERNAL TABLE');
```
* [current_schema()](https://docs.aws.amazon.com/redshift/latest/dg/c_sql-functions-leader-node.html) and [current_schemas](https://docs.aws.amazon.com/redshift/latest/dg/r_CURRENT_SCHEMAS.html) are leader node functions
* [SVV_ALL_TABLES](https://docs.aws.amazon.com/redshift/latest/dg/r_SVV_ALL_TABLES.html) is " a union of Amazon Redshift tables as shown in SVV_REDSHIFT_TABLES and the consolidated list of all external tables from all external schemas", so doesn't need to push current_schemas() to compute nodes (which would otherwise fail)
* BUT [search_path](https://docs.aws.amazon.com/redshift/latest/dg/r_search_path.html) says
> * Search paths aren't supported with external schemas and external tables. External tables must be explicitly qualified by an external schema. Datashares behave as external data and are therefore not supported as search paths.

So if we really want to support external tables from Redshift, we'll need to supply the qualified name, which in SQLA we should likely do with Table(..., schema=qualification), requiring additional changes beyond this query. I think it would be cleaner to defer this query change til that day; users will spot a problem sooner when their tables don't show up, vs breaking when we try querying them during exp design.

Thanks for the review Doug!

PR #213, opened 2026-04-09
Make table_name and primary_key required for frequentist exps by moving into DesignSpec, and similarly for PowerRequest

qixotic:
As a followup to #183, we change the API payloads for frequentist experiment creation and power checks by moving table_name and primary_key into the DesignSpec. This lets us make the fields required for frequentist experiments, be a little clearer in the API docs, and simplify some code.

FE change:  https://github.com/agency-fund/evidential-fe/pull/239


src/xngin/apiserver/routers/common_api_types.py
```diff
@@ -1128,18 +1140,7 @@ class BayesABExperimentSpec(BaseBanditExperimentSpec):
...
-        Field(description="Table name for ad-hoc power calculations. Fields are verified against the inspected table."),
-    ]
-    primary_key: Annotated[str, Field(description="Primary key field name.")]
-
-    @model_validator(mode="after")
-    def check_table_name_and_primary_key_together(self) -> Self:
-        if (self.table_name is None) != (self.primary_key is None):
-            raise ValueError("table_name and primary_key must be provided together or both omitted")
-        return self
+    design_spec: BaseFrequentistDesignSpec
```
dctaf: Interesting Pydantic question: Will BaseFrequentistDesignSpec constrain the value of .experiment_type to the union of its subclasses automatically, or will this allow values that satisfy only BaseFrequentistDesignSpec (without an .experiment_type) field to squeeze through? If it is too loose, consider a new discriminated union of the allowed leaf types. We probably do not want to ever construct a BaseFrequentistDesignSpec directly.

src/xngin/apiserver/routers/experiments/experiments_common.py
```diff
@@ -245,11 +242,9 @@ async def create_experiment_impl(
             )
 
         case ExperimentsType.FREQ_ONLINE:
-            assert request.table_name is not None
-            assert request.primary_key is not None
-            field_type_map = await fetch_fields_or_raise(
-                datasource, request.design_spec, request.table_name, request.primary_key
-            )
+            online_spec = request.design_spec
+            assert isinstance(online_spec, BaseFrequentistDesignSpec)
```
dctaf: Perhaps OnlineFrequentistExperimentSpec and  PreassignedFrequentistExperimentSpec instead of BaseFrequentistDesignSpec (here and elsewhere)?

dctaf (changes requested):

src/xngin/apiserver/routers/common_api_types.py
qixotic: Thanks! Yeah so while we can use a base class for collecting common fields shared by different models that logically derive from it, if we wish to retain the type, fields and the data added by the derived classes (which we'll always want to do), we must use a union of leaf types, else we risk pydantic scrubbing that info if it instantiates a copy with that base type.

src/xngin/apiserver/routers/experiments/experiments_common.py
qixotic: Can't use type aliases and annotated as isinstance() second arg, although a raw union assignment can. Also the union wouldn't work in a match-case: Class pattern class must not be a type alias with type parameters.

I've decided to list it out explicitly or otherwise use a match-case on the spec in question directly, so that we enumerate all the cases or otherwise acknowledge we don't care about other types with a _.

qixotic (commented):
Tacked on a little more cleanup to make the bandit set of spec types consistent. Also added another validator in https://github.com/agency-fund/evidential-be/pull/213/changes/50ba114a564ae5677d0f1625f116550756fdccb9 to check even earlier whether the pk is accidentally included by the FE in the list of strata.

dctaf (approved):

PR #215, opened 2026-04-09
Remove experiment IDs hack.

dctaf:
## Summary
Removes the unused special-case handling for fields named experiment_ids.

## What changed
- removed the experiment_ids filter validator and related docs
- removed the SQL regex special case for experiment_ids
- removed tests that only covered the hack

## Validation
- task test -- src/xngin/apiserver/routers/test_common_api_types.py
- task test -- src/xngin/apiserver/dwh/test_query_constructors.py
- task test -- src/xngin/apiserver/dwh/test_dialect_sql.py


poornimaramesh (approved):
Works, and looks good to me. Thanks!

PR #219, opened 2026-04-11
get_assignments 1M support and a bugfix in get_assignment_cmab

dctaf:
Previously, `experiments_api.get_experiment_assignments` would OOM on 1M
assignments. Now it completes in ~3s.

Also, the documented behavior of `experiments_api.get_assignment_cmab` differed
from reality. The two have now been aligned.

Also, moves a test from the common to the API level.

### DIFF
--- a/src/xngin/apiserver/routers/experiments/experiments_common_csv.py
+++ b/src/xngin/apiserver/routers/experiments/experiments_common_csv.py
@@ -192,7 +192,7 @@ async def get_experiment_assignments_impl(
                     "arm_id": arm_id,
                     "arm_name": arm_name,
                     "created_at": created_at,
-                    "strata": [],
+                    "strata": None,
                     "observed_at": observed_at,
                     "outcome": outcome,
                     "context_values": context_values,
--- a/src/xngin/apiserver/routers/experiments/experiments_common_csv.py
+++ b/src/xngin/apiserver/routers/experiments/experiments_common_csv.py
@@ -1,11 +1,13 @@
 # mypy: disable-error-code="misc"
 from collections.abc import AsyncGenerator
+from typing import cast
 
 from fastapi.responses import StreamingResponse
 from psycopg import sql
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from xngin.apiserver.exceptions_common import LateValidationError
+from xngin.apiserver.routers.common_api_types import AssignmentTypedDict, StrataTypedDict
 from xngin.apiserver.routers.common_enums import ExperimentsType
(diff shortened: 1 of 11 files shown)


src/xngin/apiserver/routers/experiments/experiments_common_csv.py
```diff
@@ -1,95 +1,201 @@
...
+        case ExperimentsType.FREQ_ONLINE.value | ExperimentsType.FREQ_PREASSIGNED.value:
+            strata_names = await _get_assignment_csv_strata_names_from_experiment(experiment)
+            select_query = _build_freq_experiment_assignments_select_query(
+                experiment.id,
+                experiment.experiment_type,
+                strata_names,
+                with_microseconds=True,
+            )
+            async for assignment in stream(xngin_session, select_query, JSON_STREAM_FETCH_SIZE_ROWS):
+                participant_id, arm_id, arm_name, created_at, *strata_values = assignment
```
poornimaramesh, line 165: nit: this seems like a risky way to unpack values from the query -- it took me a couple of minutes of scrolling back and forth to verify the order was correct.

src/xngin/apiserver/routers/experiments/experiments_common_csv.py
```diff
@@ -1,95 +1,201 @@
...
+        case ExperimentsType.MAB_ONLINE.value | ExperimentsType.CMAB_ONLINE.value:
+            select_query = _build_bandit_experiment_assignments_select_query(
+                experiment.id,
+                experiment.experiment_type,
+                with_microseconds=True,
+                include_observed_at=True,
+                include_context_vals=True,
+            )
+            async for assignment in stream(xngin_session, select_query, JSON_STREAM_FETCH_SIZE_ROWS):
+                participant_id, arm_id, arm_name, created_at, outcome, observed_at, context_values = assignment
```
poornimaramesh, line 189: same comment as on L:165

poornimaramesh (approved):
This is so awesome! -- I tried with 100k participants (I got too impatient to try and insert a million participants), but it's insane how quick the get_participants endpoint is O_O

src/xngin/apiserver/routers/experiments/experiments_common_csv.py
dctaf: I agree. I have strengthened the unit tests regarding column ordering. In the process, I found a pre-existing bug in how we represent strata when they are not in use (null vs []). To retain backwards compatibility I have retained the existing behavior and filed a bug for us to discuss later.

PR #221, opened 2026-04-14
Add cluster assignment and balance check support

mdulieu:
## Changes

**`balance.py`**
- Added `cluster_col: str | None = None` parameter to `check_balance_of_preprocessed_df()`
- When `cluster_col` is provided, uses clustered standard errors (`cov_type="cluster"`) instead of HC1. 
- `cluster_col` is excluded from covariates to avoid being used as a predictor

**`assignment.py`**
- Added `cluster_col: str | None = None` parameter to `assign_treatment_and_check_balance()`
- Added new `assign_clusters_to_arms()` function which:
  - Builds a cluster-level DataFrame with one row per cluster and its size
  - Assigns clusters to arms using `stochatreat`, stratified by cluster size buckets
  - Broadcasts arm assignments back to individual level
  - Runs a balance check at the cluster level
  - Returns `balance_result=None` when all clusters are the same size (consistent with individual path behavior when no usable stratification columns exist)

## Tests

**`test_balance.py`** — 3 new tests:
- `test_check_balance_with_cluster_col` — verifies clustered SEs differ from HC1 SEs
- `test_check_balance_with_balanced_clusters` — balanced clusters produce high p-value
- `test_check_balance_with_unbalanced_clusters` — unbalanced clusters produce low p-value

**`test_assignment.py`** — 3 new tests:
- `test_assign_clusters_equal_sizes` — equal-sized clusters produce balanced arms, no balance result
- `test_assign_clusters_unequal_sizes` — unequal clusters stratified by size, balance result present
- `test_assign_clusters_three_arms_unequal_sizes` — same but with 3 arms

## Known limitations

- Balance check currently only compares arms 0 and 1, regardless of number of arms — follow-up PR
- When `cluster_col` is provided, `stratum_cols` is ignored
- No check that arm populations meet minimum sample size requirements
Should we check that arm populations meet the minimum sample size from the power calculation after cluster assignment or should we flag when the ratio between the largest and smallest arm exceeds 1.5:1 (or another ratio)?


qixotic (approved):
thanks!

qixotic:
> No check that arm populations meet minimum sample size requirements
> Should we check that arm populations meet the minimum sample size from the power calculation after cluster assignment or should we flag when the ratio between the largest and smallest arm exceeds 1.5:1 (or another ratio)?

Rather than use a heuristic, we could have the FE highlight this in the review screen when creating a new experiment, and let the user decide to go back and re-randomize or accept the risk.

PR #222, opened 2026-04-15
Feature/top two ts

markbotterill:
Changes   
`bandit_sampling.py`
  - Added Top-Two Thompson Sampling for experiments with ≥5 arms 
  - When n_arms >= 5, choose_arm draws two independent posterior samples: the first picks a "leader" (likely best arm), the second picks a "challenger" (best arm excluding the leader). A coin flip with β=0.9 decides which to play.                                                                                                  
  - For <5 arms, regular Thompson Sampling is unchanged           
  - New functions: _top_two_beta_binomial(), _top_two_normal()                                                                                                          
  - Module-level constants TOP_TWO_MIN_ARMS = 5 and TOP_TWO_BETA = 0.9 for easy tuning - NB we can put these in a higher level confif if necessary?                                                               
  - Minor docstring cleanup in _update_arm_beta_binomial and choose_arm
 
`common_api_types.py`                                                                                                                           
  - Fixed copy-paste typo: "Prior predictive" → "Posterior predictive" in post_pred_mean and post_pred_stdev field descriptions                                         
   
Tests:                                                                                                                                                              
                                                                  
  test_bandit_sampling.py — 3 new tests:                                                                                                                                
   
  - test_top_two_returns_valid_arm_for_all_prior_types — Top-Two returns a valid arm for Beta-Binomial, Normal-Normal, and Laplace with ≥5 arms
  - test_top_two_sometimes_picks_challenger — with a dominant arm (alpha=1000), Top-Two picks challengers ≥10/200 times. Regular TS gets ~3/200 — this test would fail without Top-Two
  - test_below_threshold_barely_explores_with_dominant_arm — with <5 arms, regular TS picks the dominant arm ≥96% of the time. Confirms Top-Two is not active below threshold

Background:                                                                                                                                                          

Explored batching, forced initial sampling, burn-in periods, and Top-Two TS across Beta-Binomial, Normal-Normal, and Laplace priors. Top-Two with β=0.9 was the only approach that consistently improved best-arm identification at higher arm counts without adding infrastructure complexity. Notebook shareable on this and Notion page documenting findings [here](https://www.notion.so/agencyfund/Thompson-Sampling-Simulations-3426311d558d8074af43ed45427964ff)                    
                                                                  
Known limitations:

  - β=0.9 is hardcoded — could be made configurable per experiment in a follow-up                                                                                       
  - Top-Two threshold of 5 arms is based on simulation evidence with best=0.6, spread=0.3 — may need revisiting for very different prior configurations
  - No UI indication that Top-Two is active - it's quite an in-depth feature for users to understand and only kicks in under certain circumstances w/ limited impact -> not entirely sure we need to have the UI indicator!


src/xngin/stats/bandit_sampling.py
```diff
@@ -176,45 +186,101 @@ def choose_arm(
...
+    alphas: np.ndarray,
+    betas: np.ndarray,
+    random_state: int | None = None,
+) -> int:
+    """Top-Two Thompson Sampling for Beta-Binomial arms."""
+    rng = np.random.default_rng(random_state)
+    sample1 = rng.beta(alphas, betas)
+    leader = int(sample1.argmax())
+    sample2 = rng.beta(alphas, betas)
+    sample2[leader] = -np.inf
```
poornimaramesh: why sample twice? The reward prob for each arm is effectively independent, so you could simply sample once and take the top two instead of argmax, no?

src/xngin/stats/bandit_sampling.py
```diff
@@ -176,45 +186,101 @@ def choose_arm(
...
+    scores2 = []
+    for mu, cov in zip(mus, covariances, strict=True):
+        theta1 = rng.multivariate_normal(mean=mu, cov=cov)
+        scores1.append(link_function(context @ theta1))
+        theta2 = rng.multivariate_normal(mean=mu, cov=cov)
+        scores2.append(link_function(context @ theta2))
+
+    leader = int(np.argmax(scores1))
+    scores2[leader] = np.array(-np.inf)
+    challenger = int(np.argmax(scores2))
```
poornimaramesh: same comment as for L251-L255

src/xngin/stats/bandit_sampling.py
poornimaramesh: nit: also noting that if you switch to simply taking the top two instead of arg-maxing, you could use the `if two_sample` conditional block directly inside the _sample_normal / _sample_beta_binomial functions instead of repeating the calculations in a separate function

src/xngin/stats/bandit_sampling.py
```diff
@@ -168,6 +168,16 @@ def objective(theta: np.ndarray) -> float:
...
 # ------------- Import functions ----------------
 # --- Choose arm function ---
+# Top-Two TS configuration
+# With ≥5 arms, use Top-Two TS: draw two independent posterior samples, pick
+# between the leader (best in sample 1) and challenger (best in sample 2,
+# excluding leader) with probability β. This provides systematic exploration
+# of the most promising alternative without distorting the posterior.
+# β=0.9 means 90% of draws go to the likely best arm — low exploration cost.
+TOP_TWO_MIN_ARMS = 5  # NB can move these to config if cleaner
+TOP_TWO_BETA = 0.9
```
poornimaramesh: it would be good to discuss at some point if there's a way to allow users to configure this intuitively -- could turn it into an issue for now?

src/xngin/stats/test_bandit_sampling.py
```diff
@@ -103,3 +103,91 @@ def test_update_arm():
...
+        experiment_type=ExperimentsType.MAB_ONLINE,
+        design_spec=design_spec,
+        state=ExperimentState.COMMITTED,
+        stopped_assignments_at=None,
+        stopped_assignments_reason=None,
+    )
+    experiment = converter.get_experiment()
+    for i, arm in enumerate(experiment.arms):
+        arm.id = f"arm_{i}"
+    return experiment
```
poornimaramesh: given that `_make_many_arm_experiment` simply adds more arms to `make_createexperimentrequest_json`, I'd suggest updating the latter function to just take `num_arms` as an optional parameter and reusing `make_experiment_table` to generate a >=2 arm experiment. This way, less repeted code, and fewer instances of DesignSpec and ExperimentStorageConverter to update -- both are components that are likely to undergo changes frequently

src/xngin/stats/test_bandit_sampling.py
```diff
@@ -103,3 +103,91 @@ def test_update_arm():
...
+    Regular TS with alpha=1000 vs Beta(1,1) would pick the dominant arm
+    nearly every time. We check that at least 10/200 selections are
+    non-dominant — this would fail under regular TS.
+    """
+    experiment = _make_many_arm_experiment(6, PriorTypes.BETA, LikelihoodTypes.BERNOULLI)
+    experiment.arms[0].alpha = 1000.0
+
+    selections = [choose_arm(experiment=experiment, random_state=s).id for s in range(200)]
+    non_dominant = sum(1 for s in selections if s != "arm_0")
+    assert non_dominant >= 10, f"Expected Top-Two to explore challengers, but got {non_dominant}/200"
```
poornimaramesh: shouldn't this be 20? I'd suggest instead sticking with percentages for the test + using `pytest.approx` i.e.
```
(non_dominant / 200) == pytest.approx(0.1, rel=.01)
```

poornimaramesh: I'd also suggest using the `TOP_TWO_BETA` value here, instead of hard-coding the expected percentage

src/xngin/stats/test_bandit_sampling.py
```diff
@@ -103,3 +103,91 @@ def test_update_arm():
...
+    Regular TS with alpha=1000 vs Beta(1,1) picks the dominant arm nearly
+    every time. We check that non-dominant selections are ≤8/200 — this
+    would fail under Top-Two (which gets ~20+/200 empirically).
+    """
+    experiment = _make_many_arm_experiment(TOP_TWO_MIN_ARMS - 1, PriorTypes.BETA, LikelihoodTypes.BERNOULLI)
+    experiment.arms[0].alpha = 1000.0
+
+    selections = [choose_arm(experiment=experiment, random_state=s).id for s in range(200)]
+    non_dominant = sum(1 for s in selections if s != "arm_0")
+    assert non_dominant <= 8, f"Expected regular TS to exploit dominant arm, but got {non_dominant}/200 non-dominant"
```
poornimaramesh: same comments as on L177

src/xngin/stats/bandit_sampling.py
poornimaramesh: also, if we're being statistically rigorous, Top Two Thompson Sampling won't work the same way for contextual bandits -- guaranteeing that a challenger arm is chosen with probability 1-β _marginalized across all contexts_ is non-trivial to implement. I wouldn't worry about it right now, since no one is actually using CMABs, but maybe worth creating an issue so we can track it

src/xngin/stats/bandit_sampling.py
poornimaramesh: you could consider moving these variables to `xngin/apiserver/limits.py`

poornimaramesh (changes requested):
Thanks! This looks good to me overall, and it works as expected from testing locally. 
I've requested a few changes to the tests / implementation -- happy to discuss and/or re-review before merging.

src/xngin/stats/bandit_sampling.py
markbotterill: Screenshot 2026-04-16 at 11 45 01

In the 2020 Russo paper the algorithm is defined as above (link to paper [here](https://djrusso.github.io/docs/best_arm_identification_body.pdf)) which does two distinct samples. Logically, given that each draw is independent, P(Arm_X being 2nd best) should be === P(Arm X being the Best in a smaller draw | Arm Y has already been removed in a first draw). 

I did back this up by doing some Monte Carlo simulations last night and got identical results. As such, I think it makes sense to do just one sample. 

```
Two-sample selection distribution:
  arm_0: 89.8%
  arm_1: 3.4%
  arm_2: 3.4%
  arm_3: 3.4%
  arm_4: 0.0%

Single-sample selection distribution:
  arm_0: 89.7%
  arm_1: 3.4%
  arm_2: 3.5%
  arm_3: 3.5%
  arm_4: 0.0%
```

src/xngin/stats/test_bandit_sampling.py
markbotterill: 200 is correct - w/ Beta as 0.9 you would expect ~20 per 200 to be explored challengers. Bumped down to 10 for noise. Switching to the `TOP_TWO_BETA` though - much cleaner!

src/xngin/stats/test_bandit_sampling.py
```diff
@@ -103,3 +109,65 @@ def test_update_arm():
...
+    every time — challenger rate should be close to 0, not TOP_TWO_BETA.
+    """
+    experiment = _make_many_arm_experiment(TOP_TWO_MIN_ARMS - 1, PriorTypes.BETA, LikelihoodTypes.BERNOULLI)
+    for arm in experiment.arms:
+        arm.alpha, arm.beta = 1.0, 1.0
+    experiment.arms[0].alpha = 1000.0
+
+    selections = [choose_arm(experiment=experiment, random_state=s).id for s in range(N_DRAWS)]
+    challenger_rate = sum(1 for s in selections if s != "arm_0") / N_DRAWS
+    assert challenger_rate < EXPECTED_CHALLENGER_RATE / 2
```
poornimaramesh, line 178: nit: could you replicate these tests for Normal-Normal MABs and for CMABs?

poornimaramesh (approved):
Looks good, thanks! Approving. Could you pull latest changes from main before merging?

src/xngin/stats/test_bandit_sampling.py
markbotterill: Done!

PR #225, opened 2026-04-21
Add cluster_col support to analyze_experiment

mdulieu:
## Changes

**`analysis.py`**
- Added `cluster_col: str | None = None` parameter to `analyze_experiment()`
- When `cluster_col` is provided, uses clustered standard errors (`cov_type="cluster"`) instead of HC1
- Updated column validation to accept `cluster_col` in `assignments_df`
- `cluster_col` is excluded from metric columns to avoid being treated as an outcome
- `non_missing` mask ensures cluster groups vector aligns correctly with rows dropped due to missing values

## Tests

**`test_analysis.py`** — 5 new tests:
- `test_analysis_with_cluster_col` — clustered SEs differ from HC1 SEs; estimates are identical
- `test_analysis_cluster_col_accepts_valid_assignments` — validation accepts `assignments_df` with cluster column
- `test_analysis_with_cluster_col_unequal_sizes` — works correctly with unequal cluster sizes
- `test_analysis_with_cluster_col_missing_values` — `non_missing` mask handles missing outcomes correctly
- `test_analysis_with_cluster_col_three_arms` — works correctly with three arms

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -196,10 +196,9 @@ jobs:
           lfs: true
           persist-credentials: true
       - uses: astral-sh/setup-uv@e06108dd0aef18192324c70427afc47652e63a82 # v7.5.0
-      - uses: arduino/setup-task@b91d5d2c96a56797b48ac1e0e89220bf64044611 # v2.0.0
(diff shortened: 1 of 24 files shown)


qixotic (approved):
Rebased on main and added some commits to cleanup some tests and code a bit. LGTM, thanks!

PR #228, opened 2026-04-23
Remove unused STDDEV_POP overrides.

dctaf:
The custom stddev_pop GenericFunction existed only to translate to MSSQL's
STDEVP alias, but the codebase only supports Postgres, Redshift, and
BigQuery — all three natively support standard SQL STDDEV_POP. The override
had the side effect of converting all STDDEV_POP types to Decimal rather
than native floats.


markbotterill (approved):
LGTM! 🚀

qixotic (approved):

PR #230, opened 2026-04-23
Remove remnants of xurl.

dctaf:
In the ancient past, when Evidential was stateless, we tested the API using
static call/response test scripts and a helper called Xurl.  This PR removes
the final remnants of that behavior.


markbotterill (approved):
This looks great! I raised a micro-PR in the docs repo [here](https://github.com/agency-fund/evidential-docs/pull/31) to remove any mention of `xurl`.

PR #232, opened 2026-04-23
Allow Turn.io configuration on Evidential

poornimaramesh:
## Ticket

Tackles issue [288](https://github.com/agency-fund/evidential-sprint/issues/288)

## Related PRs
Frontend: [PR 232](https://github.com/agency-fund/evidential-fe/pull/235)

## Description
Add end points to allow users to add a Turn.io API Key, cache Journeys from a Turn.io account, configure an arm -> Journey mapping per experiment, and allow Turn.io to retrieve the required experiment configuration for routing Journeys.

- Updated tables to track Turn.io API Key (one per org) and cached Joourneys; added table to track arm -> Journeys map per experiment
- Added admin integrations endpoints for dealing with API key and Journeys
- Added public facing endpoint for Turn to retrieve a config
- Added tests for both
- Minor changes to generated API client code / app routes code to track separate set up for the integrations end points

## How has this been tested?
Run tests

## Checklist

Fill with `x` for completed.


migrations/sa_postgres/atlas.sum
```diff
@@ -62,4 +62,6 @@ h1:OsOJbKQ8J2u7Hw3pduZH6SbECaqIwyV2aR3XF64S3fY=
 20260320223010.sql h1:taOOxZD06i/Z5VQSWeutd8a91o18JItTacI8d95V9JQ=
 20260327223156.sql h1:ZS1+gQoLhXeo/MLRlOcU4kAvGGPZxz6uuDYQekc/bjw=
 20260330085015.sql h1:00EaRkus5UT71hsXqbfUt50+/d89Xtdp/Olr5SjQ94A=
-20260422163439.sql h1:aSjJr2a0pCzhd7zLbo1brfHu1aOKzIQPn4wJ1hmrKso=
+20260421135352.sql h1:cvWlPiuKH/p4w761ERYSdEd+nv6JAUnWnLgWNbhiv4k=
+20260421143753.sql h1:lpl11Mv62eQh5nCUFUNu+tsWXF2/t2gYHQbvSr9lACU=
+20260421203630.sql h1:7gzSL/WBTC7PaXZ85XhBRKRhfgC1tbeiLYxOuQPUQ5I=
```
dctaf: Process suggestion: When you're done with any changes to the tables in PR, consider squashing all the migrations into a single commit so that they can be reviewed more easily and be easier to merge against any conflicting changes on main. There is also a performance win to keeping the # of migration files to a minimum (until we checkpoint the migrations).

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
```diff
@@ -0,0 +1,325 @@
...
+
+    Creates a Turn connection for the organization if one does not yet exist, otherwise
+    overwrites the existing token. An organization has at most one Turn connection.
+    """
+    org = await get_organization_or_raise(session, user, organization_id)
+
+    turn_connection = (
+        await session.execute(select(tables.TurnConnection).where(tables.TurnConnection.organization_id == org.id))
+    ).scalar_one_or_none()
+    if turn_connection is None:
```
dctaf, line 110: PUT should be idempotent generally -- so suppose there is an infrastructure hiccup and the PUT gets retried with the same token. This will delete the ExperimentTurnConfigs unnecessarily. Consider making this idempotent with a `body.turn_api_token == turn_connection.turn_api_token` check.

.pre-commit-config.yaml
```diff
@@ -60,7 +60,7 @@ repos:
         require_serial: true
       - id: generate-typed-clients
         entry: uv run xngin-cli generate-typed-clients
-        files: (experiments_api|admin_api)\.py$
+        files: (experiments_api|admin_api|admin_integration_api|integration_api)\.py$
```
dctaf: `integration_api` => `integrations_api`; but, more broadly, should we make them consistent?

src/xngin/apiserver/routers/admin_integrations/admin_integration_api.py
```diff
@@ -0,0 +1,325 @@
...
+        and turn_connection.cached_journeys_updated_at is not None
+        and now - turn_connection.cached_journeys_updated_at < timedelta(seconds=TURN_JOURNEYS_CACHE_TTL_SECONDS)
+    ):
+        return GetTurnJourneysResponse(journeys=turn_connection.cached_journeys)
+
+    async with httpx.AsyncClient(timeout=TURN_REQUEST_TIMEOUT_SECONDS) as client:
+        response = await client.get(
+            "https://whatsapp.turn.io/v1/stacks",
+            headers={"Authorization": f"Bearer {turn_connection.get_turn_api_token()}"},
+        )
```
dctaf: This might raise httpx.RequestError if there are failures unrelated to status (e.g. DNS error). That will generate a 500 for the user rather than a helpful error message. Also, raise_for_status is quite tolerant about error codes -- consider testing for the exact error code the API has committed to instead, ensuring we don't misinterpret the response body.

Consider something like this:

```python
try:
  async with ... as client:
    response = await client...
except httpx.RequestError...:
  ... raise 502
if response.status != 200:
  raise 
```

src/xngin/apiserver/routers/admin_integrations/admin_integration_api.py
```diff
@@ -0,0 +1,325 @@
...
+
+    try:
+        response.raise_for_status()
+    except httpx.HTTPStatusError as exc:
+        logger.error(f"Error fetching Turn.io journeys: {exc.response.status_code} - {exc.response.text}")
+        raise HTTPException(
+            status_code=exc.response.status_code,
```
dctaf: If the client benefits from knowing the specific error code that Turn returns then let's represent that in GetTurnJourneysResponse instead (for example, in a `.turn_status_code` field). Conflating our response codes with Turn's will make elegant API clients more difficult to reason about. While our own API response codes are somewhat inconsistent, adding a 3rd parties codes into the mix will make it even less consistent.

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
```diff
@@ -0,0 +1,325 @@
...
+    organization_id: str,
+    session: Annotated[AsyncSession, Depends(xngin_db_session)],
+    user: Annotated[tables.User, Depends(require_user_from_token)],
+    allow_missing: Annotated[
+        bool,
+        Query(description="If true, return a 204 even if the resource does not exist."),
+    ] = False,
+):
+    """Removes an organization's Turn.io connection."""
+    resource_query = select(tables.TurnConnection).where(tables.TurnConnection.organization_id == organization_id)
```
dctaf, line 192: TurnConnection doesn't have a cascade delete relationship with ExperimentTurnConfig. Is this intentional?

src/xngin/apiserver/routers/admin_integrations/admin_integration_api.py
```diff
@@ -0,0 +1,325 @@
...
+    ds = await get_datasource_or_raise(session, user, datasource_id)
+    experiment = await get_experiment_via_ds_or_raise(session, ds, experiment_id)
+    turn_connection = (
+        await session.execute(
+            select(tables.TurnConnection).where(tables.TurnConnection.organization_id == ds.organization_id)
+        )
+    ).scalar_one_or_none()
+    if not turn_connection:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
```
dctaf: Usually 400 means that the request will /never/ succeed; but in this case, the request will succeed if the TurnConnection is configured. We're not too disciplined about this but if you want the client to be able to distinguish between these error conditions, consider alternate error codes.

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api_types.py
```diff
@@ -0,0 +1,65 @@
...
+
+from pydantic import Field
+
+from xngin.apiserver.routers.admin.admin_api_types import AdminApiBaseModel
+
+
+class SetConnectionToTurnRequest(AdminApiBaseModel):
+    """Request to create or update an organization's Turn.io connection."""
+
+    turn_api_token: Annotated[
```
dctaf, line 11: Security: Validate that the turn_api_token does not contain newlines. This mitigates HTTP request header injection vulnerabilities.

src/xngin/apiserver/routers/admin_integrations/test_admin_integration.py
```diff
@@ -0,0 +1,334 @@
...
+from xngin.apiserver.routers.experiments.experiments_common import fetch_fields_or_raise
+from xngin.apiserver.sqla import tables
+from xngin.apiserver.storage.storage_format_converters import ExperimentStorageConverter
+from xngin.apiserver.testing.admin_api_client import AdminAPIClient
+from xngin.apiserver.testing.admin_integrations_api_client import AdminIntegrationsAPIClient
+from xngin.apiserver.testing.testing_dwh_def import TESTING_DWH_PARTICIPANT_DEF
+
+
+@pytest.fixture(name="testing_experiment")
+async def fixture_testing_experiment(xngin_session: AsyncSession, testing_datasource) -> tables.Experiment:
```
dctaf: Consider replacing this with an implementation that uses the AdminAPIClient instead of direct database access.

src/xngin/apiserver/routers/integrations/test_integrations_api.py
```diff
@@ -0,0 +1,120 @@
...
+from xngin.apiserver.routers.experiments.experiments_common import fetch_fields_or_raise
+from xngin.apiserver.sqla import tables
+from xngin.apiserver.storage.storage_format_converters import ExperimentStorageConverter
+from xngin.apiserver.testing.admin_integrations_api_client import AdminIntegrationsAPIClient
+from xngin.apiserver.testing.integrations_api_client import IntegrationsAPIClient
+from xngin.apiserver.testing.testing_dwh_def import TESTING_DWH_PARTICIPANT_DEF
+
+
+@pytest.fixture(name="testing_experiment")
+async def fixture_testing_experiment(xngin_session: AsyncSession, testing_datasource) -> tables.Experiment:
```
dctaf: Consider using AdminAPIClient instead of direct database access.

dctaf (approved):
Love it! Address the issues below and feel free to merge at your discretion.

migrations/sa_postgres/atlas.sum
poornimaramesh: done, thanks!

.pre-commit-config.yaml
poornimaramesh: I've renamed everything to `*integrations*`, thanks!

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
poornimaramesh: thanks for catching, I've fixed this

src/xngin/apiserver/routers/admin_integrations/admin_integration_api.py
poornimaramesh: I am not sure that a client will want to know the precise nature of the error from Turn.io (most people use the UI afaik, and not the API). That said, I've settled for returning both the original status code and error message in the details of the exception we raise.

For now, I return a 409 in case of 401 / 403 errors from Turn.io. And I don't differentiate between connection, timeout errors and non-200 errors, and simply handle it as a 502 (Claude's recommendation).

I've made [an issue](https://github.com/agency-fund/evidential-sprint/issues/309) to track this though, and we can revisit it if/when we see use cases that need a more elegant solution?

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
poornimaramesh: It should! THanks for catching, I've fixed this and added tests

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api_types.py
poornimaramesh: Done, thanks! Didn't know this was a security risk, so I ended up going to down a rabbit hole -_-

src/xngin/apiserver/routers/admin_integrations/test_admin_integration.py
poornimaramesh: Done, thanks!

src/xngin/apiserver/routers/admin_integrations/admin_integration_api.py
poornimaramesh: Updated it to a 409 error (Claude' suggestion), but we can revisit if needed with [Issue 309](https://github.com/agency-fund/evidential-sprint/issues/309)

src/xngin/apiserver/routers/integrations/test_integrations_api.py
poornimaramesh: done, thanks

src/xngin/apiserver/routers/admin_integrations/admin_integration_api.py
poornimaramesh: done, thanks!

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
```diff
@@ -185,21 +207,36 @@ async def get_organization_turn_journeys(
...
+        logger.error(f"Error fetching Turn.io journeys: {exc}")
         raise HTTPException(
-            status_code=exc.response.status_code,
-            detail=f"Error fetching Turn.io journeys: {exc.response.text}",
+            status_code=status.HTTP_502_BAD_GATEWAY,
+            detail=f"Failed to reach Turn.io API to fetch journeys. Details: {exc}",
         ) from exc
 
+    if response.status_code in {401, 403}:
+        logger.error("Unauthorized when fetching Turn.io journeys. API token may be invalid.")
```
dctaf: If you feel like exploring logging further, consider annotating the log lines with `with logger.contextualize(turn_experiment_id=X):` so that you can extract per-connection debugging information from the logs more easily. We do this in the snapshotter and it has occasionally been helpful for debugging.

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api_types.py
```diff
@@ -18,6 +18,14 @@ class SetConnectionToTurnRequest(AdminApiBaseModel):
         ),
     ]
 
+    @field_validator("turn_api_token")
+    @classmethod
+    def validate_turn_api_token(cls, v: str) -> str:
+        "Check for newline or carriage return characters to prevent HTTP header injection."
```
dctaf: Method comments should be triple-quoted.

src/xngin/apiserver/routers/integrations/test_integrations_api.py
```diff
@@ -2,119 +2,106 @@
...
 @pytest.fixture(name="arm_to_journey_mapping")
 def fixture_arm_to_journey_mapping(
+    aclient: AdminAPIClient,
     iaclient: AdminIntegrationsAPIClient,
-    testing_datasource,
-    testing_experiment: tables.Experiment,
-) -> Generator[dict[str, str]]:
+    testing_datasource: DatasourceMetadata,
+    testing_design_spec: MABExperimentSpec,
+) -> Generator[tuple[dict[str, str], str]]:
```
dctaf: Nit: To improve readability, I'd use a dataclass.

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
```diff
@@ -154,6 +160,22 @@ async def delete_turn_connection_from_organization(
...
+            .join(tables.Datasource, tables.Experiment.datasource_id == tables.Datasource.id)
+            .where(tables.Datasource.organization_id == organization_id)
+        )
+    )
+    await handle_delete(
+        session,
+        True,
+        authz.is_user_authorized_on_organization(user, organization_id),
+        mapping_query,
+    )
```
dctaf: I'm suspicious that this probably does not work. handle_delete() was implemented assuming that the resource would be a single row and uses `.scalar_one_or_none()`, so passing in a query that can return multiple rows is probably going to fail break something. Also, calling handle_delete() more than once in a single endpoint might have weird implications for the AIP-135 style of DELETE call (haven't verified this though).

This might be tighter: `handle_delete` supports a `deleter` argument; perhaps we should use that instead? e.g. something like this:

```python
    async def _delete_turn_connection_and_mappings(
        session: AsyncSession,
        turn_connection: tables.TurnConnection,
    ) -> None:
        await session.delete(... mappings ...)
        await session.delete(turn_connection)
    ...
    resource_query = select(tables.TurnConnection).where(tables.TurnConnection.organization_id == organization_id)
    response = await handle_delete(
        session,
        allow_missing,
        authz.is_user_authorized_on_organization(user, organization_id),
        resource_query,
        deleter=_delete_turn_connection_and_mappings,
    )

```

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
dctaf, line 94: Optional, if ya want a cleaner OpenAPI spec and possibly better generated clients (and similar for the other method):

```
TURN_JOURNEY_MAPPING_RESPONSES: dict[str | int, dict[str, Any]] = {
    "409": {
        "model": HTTPExceptionError,
        "description": "A Turn.io connection must be configured before journey mappings can be saved.",
    },
}
@router.put(
    "/integrations/turn-connection/{organization_id}",
    status_code=status.HTTP_204_NO_CONTENT,
    responses=TURN_JOURNEY_MAPPING_RESPONSES)
```

dctaf (approved):
Approved. Excited! Most of my comments are minor and could be ignored or deferred, but I do recommend at least these two changes:

1. One more place to reference the new APIs: We use `[tool.ruff.lint.flake8-import-conventions.extend-aliases]` to have Rust force imports of the various API types files to have consistent aliases that can help readability when navigating files that import types from multiple APIs. I recommend adding the new types there for consistency.
2. Use `handle_delete(deleter=...)` instead of chaining the `handle_delete` calls.

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
poornimaramesh: added, thanks! I hadn't noticed the contextualization with the snapshotter.

src/xngin/apiserver/routers/integrations/test_integrations_api.py
poornimaramesh: fixed, thanks

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
poornimaramesh: switched to a deleter, thanks!

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api_types.py
poornimaramesh: fixed, thanks!

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
poornimaramesh: Thanks! Added response mappings now.

PR #233, opened 2026-04-23
Additional bandit analysis tests

dctaf:
These tests improve coverage of corner cases in bandit analysis and will serve to ensure that upcoming performance-related changes do not alter user-visible behavior of the API.

### DIFF
--- a/src/xngin/apiserver/snapshots/test_snapshotter.py
+++ b/src/xngin/apiserver/snapshots/test_snapshotter.py
@@ -1,4 +1,5 @@
 import asyncio
+import math
 import warnings
 from datetime import UTC, datetime, timedelta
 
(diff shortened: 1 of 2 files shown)


src/xngin/apiserver/snapshots/test_snapshotter.py
```diff
@@ -599,21 +666,112 @@ async def test_create_snapshot_cmab_with_zero_draws_succeeds(
...
+        assert assignment_response.status == 200, assignment_response.data
+        if outcome is not None:
+            eclient.update_bandit_arm_with_participant_outcome(
+                api_key=testing_datasource.key,
+                body=UpdateBanditArmOutcomeRequest(outcome=outcome),
+                experiment_id=experiment_id,
+                participant_id=participant_id,
+            )
+
+    expected_contexts = [1.0 if context.value_type == ContextType.BINARY else 7.0 / 3.0 for context in sorted_contexts]
```
markbotterill: Nit: Can we we add a one-liner comment just explaining the logic here? e.g. # BINARY context uses rounded mean of assignments; REAL_VALUED uses mean

markbotterill: Just makes it a bit easier to realize where the 7.0 comes from. Additionally, we could write that 7.0 as (1 + 4 + 2)

markbotterill: Same for the binary: BINARY (1.0, 0.0, 1.0) → mean = 0.667 → ceil of (-0.5 - 0.667)=0.167 -> 1. Just makes scanning a tiny bit easier!

markbotterill (approved):
Looks great! Just added a tiny comment on test setup readability :)

src/xngin/apiserver/snapshots/test_snapshotter.py
dctaf: Done. Thank you!

PR #234, opened 2026-04-23
Improve performance of bandit analysis with large draw counts.

dctaf:
1. Adds performance benchmarks for bandit analyses. Results shared separately.
2. Replaces all the Experiment.draws preloads with on-demand aggregate queries.

The tests in #233 pass against this implementation.


src/xngin/apiserver/snapshots/snapshotter.py
```diff
@@ -161,33 +159,30 @@ async def _query_dwh_for_snapshot_data(
...
-            draws = await experiment.awaitable_attrs.draws
+        if experiment_type.is_cmab():
             contexts = await experiment.awaitable_attrs.contexts
+            sorted_contexts = sorted(contexts, key=lambda c: c.id)
+            # draw.context_vals were already sorted corresponding to the sorted_contexts
+            # ordering when the assignment was made.
+            mean_context_vals = await _mean_context_vals_from_draws(session, experiment.id, len(contexts))
+            context_vals = [
+                abs(float(np.ceil(m - 0.5))) if context.value_type == ContextType.BINARY else m
+                for m, context in zip(mean_context_vals, sorted_contexts, strict=False)
```
markbotterill: QQ: Why do we do `strict=False` here? Could we ever have different lengths?

src/xngin/apiserver/routers/experiments/experiments_common.py
```diff
@@ -1161,26 +1162,52 @@ async def read_assignments_efficiently(xngin_session: AsyncSession, experiment_i
...
+                func.coalesce(func.stddev_pop(tables.Draw.outcome), 0.0),
+            ).where(
+                tables.Draw.experiment_id == experiment_id,
+                tables.Draw.outcome.is_not(None),
+            )
+        )
+    ).one()
+    return DrawsOutcomeAggregates(
+        n_outcomes=n_outcomes,
+        outcome_std_dev=std_dev if n_outcomes > 1 else 0.0,
```
markbotterill: `func.stddev_pop(tables.Draw.outcome), 0.0)` -> when n_outcomes = 0 we get 0.0 thanks to COALESCE.

STDDEV_POP = 0 when n = 1 since population = 1 and there are no errors per se.

As such, the guard on this line never gets triggered AFAIK? Do we need it here?

markbotterill (approved):
Image

Results replicated on my end, too!

Added some small questions and clarifications but giving provisional approval!

src/xngin/apiserver/routers/experiments/experiments_common.py
dctaf: Good catch. This was needless preservation of existing syntax that is obsolete in the SQL rewrite.  Fixed.

src/xngin/apiserver/snapshots/snapshotter.py
dctaf: Good catch. create_assignment_for_participant enforces an invariant at write time that the lengths must match. Just to be safe, I've queried the production database [^1] to confirm that we don't have any data that violates the invariant. I've switched this to `strict=True`.

[^1]: Thanks to GPT-5.5 for writing the query.

PR #236, opened 2026-04-27
Add a test-only task to taskfile

qixotic:
`test-only` skips the `bootstrap-dwh-database` dependency. 
(Also makes it easier to run tests within a sandboxed agent, in conjunction with a rule that says  ``` Run tests with `UV_CACHE_DIR="${PWD}/.uv-cache" task test-only` ```, as it won't try to reach out of the current working dir to access a global uv cache nor require docker permissions to set up the db.)


dctaf (approved):
Consider removing the test-psycopg2 task while you're editing this file.

qixotic:
> Consider removing the test-psycopg2 task while you're editing this file.

Sure; we can test that config with existing flags for a redshift dwh, and we don't actually ever want to use that driver with our app db either.

PR #237, opened 2026-04-28
Dwh detection consolidation and add a redshift-specific clustered_dwh.redshift.ddl

qixotic:
## Description

Consolidate logic around detection of a dwh backend type (bigquery, postgres, redshift) into dwh_utils.py, along with the small SQLA engine setup workaround for redshift.

## Related PRs

#211 

## How has this been tested?

Ran `XNGIN_QUERIES_TEST_URI="postgresql+psycopg2://....redshift-serverless.amazonaws.com:5439/dev" task test`, all passing.

Also created another table under its own schema: `AWS_PROFILE=...  uv run xngin-cli create-testing-dwh   --dsn="postgresql+psycopg2://...redshift-serverless.amazonaws.com:5439/dev" --create-db --bucket=evi-redshift-test-data --iam-role="arn:aws:iam::..." --src src/xngin/apiserver/testdata/clustered_dwh.csv.zst --table-name clustered_dwh --schema-name myschema`
Manually configured the datasource on localhost:3000 with search path = 'myschema' only, confirmed that creation of a new experiment only sees my clustered_dwh and created successfully.


src/xngin/apiserver/dwh/dwh_utils.py
```diff
@@ -0,0 +1,31 @@
+from sqlalchemy import URL, Engine
+
+REDSHIFT_HOSTNAME_SUFFIXES = ("redshift.amazonaws.com", "redshift-serverless.amazonaws.com")
+
+
+def is_redshift(host_or_url: str | URL) -> bool:
+    """Returns true iff the hostname string or URL indicates that this is connecting to Redshift."""
+    if isinstance(host_or_url, str):
+        return host_or_url.endswith(REDSHIFT_HOSTNAME_SUFFIXES)
```
dctaf, line 13: Note for the future: This will break when customers use Redshift's [custom domains](https://docs.aws.amazon.com/redshift/latest/mgmt/connecting-connection-CNAME-create-custom-domain.html) feature.

src/xngin/apiserver/dwh/test_query_constructors.py
```diff
@@ -350,10 +351,11 @@ def test_relations(testcase, queries_dwh_session, shared_sample_tables):
     assert list(sorted([r.id for r in query_results])) == list(sorted(r.id for r in testcase.matches)), testcase
 
 
-def _datatype_to_sqlalchemy_type(data_type: DataType):
+def _datatype_to_sqlalchemy_type(data_type: DataType, is_redshift: bool):
     """Maps DataType enum to generic camel-case SQLAlchemy column type. Helper to create tables for filter tests."""
-    # DDL for sqlalchemy.types.Uuid is not supported by sqlalchemy-bigquery (falls back to invalid CHAR(32)).
-    my_uuid_type: Uuid = Uuid().with_variant(String(), "bigquery")
+    # DDL for sqlalchemy.types.Uuid is not supported by sqlalchemy-bigquery (falls back to invalid CHAR(32)),
+    # nor by Redshift, which uses the"postgresql" dialect.
```
dctaf:
```suggestion
    # nor by Redshift, which uses the "postgresql" dialect.
```

src/xngin/cli/main.py
```diff
@@ -267,9 +270,10 @@ def read_csv():
         return pd_read_csv(src, nrows=nrows)
 
     def drop_and_create(cur, create_table_ddl: str):
-        cur.execute(drop_table_ddl)
+        # Guard against trying to drop from a schema that doesn't even exist yet.
```
dctaf: Nit: drop comment

src/xngin/apiserver/conftest.py
```diff
@@ -159,16 +160,18 @@ def disable_safe_resolve_check():
 
 def get_test_uri_info(connection_uri: str) -> TestUriInfo:
     """Returns a TestUriInfo dataclass about a test database given its connection_uri."""
-    if connection_uri.startswith("bigquery"):
+    parsed_url = make_url(connection_uri)
```
dctaf, line 166: This will raise sqlalchemy.exc.ArgumentError if the connection_uri is empty string (as it might be in certain test environments without a test database). This will cause print_database_env_vars to behave differently than it does today, because it only ignores ValueError. We never run tests without a database, so I think the best path forward right now is to remove print_database_env_vars, as it doesn't add much value.

dctaf (changes requested):

src/xngin/apiserver/conftest.py
dctaf: Related: The CI failure :) https://github.com/agency-fund/evidential-be/actions/runs/25038952750/job/73337180220?pr=237#step:9:523

dctaf:
Thanks for the cleanup!

src/xngin/apiserver/conftest.py
qixotic: Thanks! Updated to raise a ValueError when empty for now; I'd like to keep print_database_env_vars for now since I've found it helpful in ad-hoc debugging in the past when messing with sourcing different .env combinations

dctaf (approved):

PR #241, opened 2026-04-30
raise error when the unique identifier is also used as a strata column

qixotic:
## Ticket

Quick fix to an error that can arise e.g. if someone clicks "Add All" under Strata in the FE, as we should not be allowing someone to include the primary key (unique identifier) to the list of strata.

## Description

This raises an error with a clearer message to the user as to why they are unable to move on beyond the power check.

## Future Tasks (optional)

Move the validation further up the chain into the design spec as part of #213.
FE can also explicitly exclude the known unique identifier column from the list. (More generally we could remove any other known unique id columns, if we knew of them in advance.)

## How has this been tested?

unittest and manual reproduction

### DIFF
--- a/src/xngin/stats/assignment.py
+++ b/src/xngin/stats/assignment.py
@@ -11,7 +11,7 @@ from xngin.stats.balance import (
     preprocess_for_balance_and_stratification,
     restore_original_numeric_columns,
 )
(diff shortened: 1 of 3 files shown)


poornimaramesh (approved):
Confirmed that this fixes the dtype error. Test pass, looks good to me, thanks!

PR #243, opened 2026-05-05
Data fix & reapplication of PR #213 (re: Make table_name and primary_key required for frequentist exps)

qixotic:
## Description

An update to the reverted #213. This reapplies PR 213 after I've cleaned up some stale test data in the db via API, and also makes it easier to see problematic ids in raised errors from caabd74d30097055512c43f4f9a53fdbb584776f.

Here's a diff with the previous merge commit: https://github.com/agency-fund/evidential-be/compare/a0f8f9b73d927346b54bf45a171eac5150c6c586...213_redo?expand=1

FE PR: https://github.com/agency-fund/evidential-fe/pull/240

## How has this been tested?

Identified problematic rows:
```
SELECT
    e.experiment_type,
    e.state,
    ef.field_name AS pk,
    e.datasource_table AS ds_table,
    e.created_at,
    e.id AS eid,
    o.name as oname,
    e.name as ename
FROM organizations o
JOIN datasources d on (d.organization_id = o.id)
JOIN experiments e on (e.datasource_id = d.id)
LEFT JOIN experiment_fields ef
    ON ef.experiment_id = e.id
    AND ef.is_unique_id = TRUE
WHERE
    experiment_type in ('freq_online', 'freq_preassigned')
    AND ef.field_name is NULL
ORDER BY created_at asc;
```

Re-confirmed that the following query hits only the 3 rows identified earlier as problematic:
```SELECT * FROM experiments e
WHERE e.experiment_type IN ('freq_online', 'freq_preassigned')
  AND e.state = 'assigned'
  AND e.created_at < '2026-01-01'
  AND datasource_table is NULL;
  ```
Data erased from the ds via API.

### DIFF
--- a/src/xngin/apiserver/conftest.py
+++ b/src/xngin/apiserver/conftest.py
@@ -519,8 +519,10 @@ def expect_status_code(
             f"Expected '{detail_eq}' to be one of the .msg fields in the response: {http_response.content}"
         )
     if detail_contains is not None:
-        assert any(detail_contains in msg for msg in match._detail_messages()), (
-            f"Expected '{detail_eq}' to be in one of the .msg fields in the response: {http_response.content}"
+        messages = match._detail_messages()
+        assert any(detail_contains in msg for msg in messages), (
+            f"Expected '{detail_contains}' to be in one of the .msg fields in the response: {http_response.content}"
+            f"\nmsg list (length {len(messages)}): \n\t{'\n\t'.join(messages)}"
         )
     if text is not None:
         assert match._has_text(text), f"Expected '{text}' to be in the response: {http_response.content}"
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

**Review by poornimaramesh — APPROVED** [review]
Tests pass, stuff works, approving!

File: migrations/sa_postgres/20260505060605.sql [hunk 1]
```diff
@@ -0,0 +1,7 @@
+-- one-off update to fix our test org with very old outdated data.
+UPDATE experiments e
+SET state = 'abandoned'
+WHERE e.experiment_type IN ('freq_online', 'freq_preassigned')
```
dctaf: The comment suggests this applies only to an organization we use for testing. Instead of a migration, can we just delete the bad experiments using the UI?

If not, then consider constraining this UPDATE to specific organization or datasource IDs.

dctaf (changes requested):

dctaf (approved):

PR #248, opened 2026-05-13
Persist is_cluster_key on ExperimentFields storage; retrievable w/ Experiments.cluster_key_field() (follow-up to #163)

mdulieu:
## Summary

PR #163 added `cluster_column` to `BaseFrequentistDesignSpec` so the API accepts cluster randomization parameters, but `storage_format_converters.py` dropped the field on save. As a result `design_spec.cluster_column` was missing from GET responses, and the frontend couldn't detect cluster experiments without falling back to reading the stored `power_analyses` JSON blob.

This PR adds the storage column and migration so cluster experiments round-trip cleanly.

## Changes

- **`src/xngin/apiserver/sqla/tables.py`** — Add nullable `cluster_column: Mapped[str | None]` to the `Experiment` model.
- **`src/xngin/apiserver/storage/storage_format_converters.py`** — Persist `cluster_column` from the design spec in `from_components`; re-attach it on the way back in `get_design_spec`.
- **`migrations/sa_postgres/20260512230000.sql`** — Atlas migration adding the column.
- **`test_admin_api.py`** — Update the two cluster power_check tests to the post-#163 schema (`table_name`/`primary_key` on the design spec, not on `PowerRequest`). These were already broken at HEAD after the merge from `main`.

## Dependencies

Depends on PR #163 — uses `design_spec.cluster_column` and the `PowerRequest`/`BaseFrequentistDesignSpec` shape introduced there. The test fix in this PR also unblocks #163's mypy hook.

## Testing

- Local round-trip: created a cluster experiment in the FE, verified `design_spec.cluster_column` is present in the GET response and survives save/reload.
- `task hash-migrations` + `task apply-migrations` applied cleanly on a fresh dev DB.
- Pre-commit hooks (ruff format, ruff check, mypy) pass.


src/xngin/apiserver/sqla/tables.py
```diff
@@ -484,6 +484,9 @@ class Experiment(Base):
     power: Mapped[float | None] = mapped_column()
     alpha: Mapped[float | None] = mapped_column()
     fstat_thresh: Mapped[float | None] = mapped_column()
+    # Column on the underlying datasource table that identifies each cluster
+    # for cluster-randomized experiments. None for non-cluster experiments.
+    cluster_column: Mapped[str | None] = mapped_column(String(255))
```
qixotic: We shouldn't store this here; we not too long ago moved to a more elaborate representation of the dwh fields used by an experiment: see the table `ExperimentField`. 

For the purpose of storing this information about which field was used as the clustering key, cluster_column's use is pretty similar to how we store the primary_key info, so you can follow the pattern established by the field backing that, i.e. add a bool is_cluster_key, and make sure it's set properly and read back correctly in the storage_format_converters.py.

BTW, I will update #163 to change `cluster_column` to `cluster_key` to be a little more dwh-agnostic-sounding, and consistent with the `primary_key` name in the `BaseFrequentistDesignSpec`.

src/xngin/stats/cluster_power.py
```diff
@@ -211,7 +211,17 @@ def solve_for_mde_cluster(
     power: float = 0.8,
     alpha: float = 0.05,
 ) -> MetricPowerAnalysis:
-    """Calculate the minimum detectable effect (MDE) for a cluster-randomized design."""
+    """Calculate the minimum detectable effect (MDE) for a cluster-randomized design.
```
qixotic: Let's move this commit 719128c0890645a348ec69f509ec152971fe31b8 into a separate PR since it looks independent from the cluster_key persistence change. Please add appropriate tests as part of the change, thanks!

qixotic (changes requested):

src/xngin/apiserver/sqla/tables.py
mdulieu: Reworked to follow the primary_key pattern. Added is_cluster_key on ExperimentField, set it in _set_experiment_fields_from_design_spec, and read it back in get_design_spec via a new cluster_key_field() helper on Experiment. New round-trip test added in test_admin_api.py. Force-pushed in 59d965b8.

qixotic (approved):
lgtm, thanks!

PR #251, opened 2026-05-18
Offload the numerically intensive convert_arm_weights to a thread

markbotterill:
Reviewer: @poornimaramesh 

Addresses issue [here](https://github.com/orgs/agency-fund/projects/8/views/1?pane=issue&itemId=170594544&issue=agency-fund%7Cevidential-sprint%7C285) and issue listed here on Sprint [board](https://github.com/agency-fund/evidential-sprint/issues/285).

NB @dctaf is right that - in situations with many arms - this could get intensive. However, it only runs at experiment creation time - even at 1M user scale I would hope we wouldn't have 1M simultaneous experiment creations.


poornimaramesh (approved):
Works, tests pass. Approving

PR #252, opened 2026-05-18
Populate cluster fields in MDE-mode solve_for_mde_cluster

mdulieu:
Split out of #248 per @qixotic's review feedback (adds tests too).

## Summary

Extend the MDE-mode response from `solve_for_mde_cluster` to include `num_clusters_total`, `clusters_per_arm`, `n_per_arm`, `design_effect`, and `effective_sample_size`. Previously these stayed `None` when `desired_n` was provided, causing saved cluster experiments to show stale cluster counts from the original recommended-size power check instead of counts that match the committed sample size.

## Tests

- `test_solve_for_mde_cluster_populates_all_fields` — regression test for all five newly-populated fields, plus the updated message.
- `test_solve_for_mde_cluster_respects_arm_weights` — verifies the per-arm split honors `arm_weights`.
- `test_solve_for_mde_cluster_unbalanced_has_larger_mde` — regression test for the `arm_weights` bonus fix below.

## Bonus fix

`solve_for_mde_cluster_impl` previously declared `arm_weights` but never forwarded it to `solve_for_mde_individual_impl`. Unbalanced cluster MDE designs were silently using balanced-arm math as a result. Fixed by passing the argument through.

Covered by `test_solve_for_mde_cluster_unbalanced_has_larger_mde`, which asserts an unbalanced 3:1 split yields a larger MDE than a balanced split with the same total `desired_n`.

### DIFF
--- a/src/xngin/stats/cluster_power.py
+++ b/src/xngin/stats/cluster_power.py
@@ -218,6 +218,8 @@ def solve_for_mde_cluster(
 ) -> MetricPowerAnalysis:
     """Calculate the minimum detectable effect (MDE) for a cluster-randomized design."""
 
+    assert metric.icc is not None and metric.avg_cluster_size is not None and metric.cv is not None
+
     target_possible, pct_change_possible = solve_for_mde_cluster_impl(
         metric=metric,
         desired_n=desired_n,
@@ -227,12 +229,39 @@ def solve_for_mde_cluster(
         power=power,
     )
 
+    # Compute DEFF, effective_n and numbers of clusters to add to MetricPowerAnalysis.
+    if arm_weights is None:
+        arm_probs = [1.0 / n_arms] * n_arms
+    else:
+        total_weight = sum(arm_weights)
+        arm_probs = [w / total_weight for w in arm_weights]
+
+    deff = calculate_design_effect(metric.icc, metric.avg_cluster_size, metric.cv)
+
+    clusters_per_arm_list: list[int] = []
+    n_per_arm_list: list[int] = []
+    for prob in arm_probs:
+        n_this_arm = desired_n * prob
+        clusters_this_arm = math.ceil(n_this_arm / metric.avg_cluster_size)
+        clusters_per_arm_list.append(clusters_this_arm)
+        n_per_arm_list.append(round(n_this_arm))
+
+    clusters_total = sum(clusters_per_arm_list)
+    effective_n = calculate_effective_sample_size(desired_n, deff)
+
     # Build response object for MDE calculation
-    analysis = MetricPowerAnalysis(metric_spec=metric)
-    analysis.target_n = desired_n
-    analysis.target_possible = target_possible
-    analysis.pct_change_possible = pct_change_possible
-    analysis.sufficient_n = None  # Not applicable in MDE mode
+    analysis = MetricPowerAnalysis(
+        metric_spec=metric,
+        target_n=desired_n,
+        target_possible=target_possible,
+        pct_change_possible=pct_change_possible,
+        sufficient_n=None,  # Not applicable in MDE mode
+        num_clusters_total=clusters_total,
+        clusters_per_arm=clusters_per_arm_list,
+        n_per_arm=n_per_arm_list,
+        design_effect=deff,
+        effective_sample_size=effective_n,
+    )
 
     # Create message
     assert metric.metric_baseline is not None
@@ -240,11 +269,13 @@ def solve_for_mde_cluster(
         "desired_n": desired_n,
         "metric_baseline": round(metric.metric_baseline, 4),
         "target_possible": round(target_possible, 4),
+        "num_clusters_total": clusters_total,
     }
     msg_type = MetricPowerAnalysisMessageType.SUFFICIENT
     msg_body = (
-        "With a desired sample size of {desired_n} units and a metric baseline of "  # noqa: RUF027
-        "{metric_baseline}, the minimum detectable effect (MDE) is {target_possible}."
+        "With a desired sample size of {desired_n} units across {num_clusters_total} "
+        "clusters and a metric baseline of {metric_baseline}, the minimum detectable "
+        "effect (MDE) is {target_possible}."  # noqa: RUF027
     )
     analysis.msg = MetricPowerAnalysisMessage(
         type=msg_type,
--- a/src/xngin/stats/cluster_power.py
+++ b/src/xngin/stats/cluster_power.py
@@ -2,6 +2,7 @@
 Power analysis for cluster-randomized designs.
 """
 
+import dataclasses
 import math
 
 from xngin.apiserver.routers.common_api_types import (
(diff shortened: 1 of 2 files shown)


src/xngin/stats/cluster_power.py
```diff
@@ -227,24 +229,53 @@ def solve_for_mde_cluster(
         power=power,
     )
 
+    # Compute DEFF, effective_n and numbers of clusters to add to MetricPowerAnalysis.
+    if arm_weights is None:
```
qixotic: Please extract the arm_probs calculation into a little helper instead of duplicating the calc here and `solve_for_sample_size_cluster`.

src/xngin/stats/cluster_power.py
```diff
@@ -227,24 +229,53 @@ def solve_for_mde_cluster(
...
+        arm_probs = [1.0 / n_arms] * n_arms
+    else:
+        total_weight = sum(arm_weights)
+        arm_probs = [w / total_weight for w in arm_weights]
+
+    deff = calculate_design_effect(metric.icc, metric.avg_cluster_size, metric.cv)
```
qixotic: deff and effective_n are already calculated in `solve_for_mde_cluster_impl`, so how about returning those as well? (In the solve_for_sample_size_cluster case, you would just ignore them.) Given the multiple values, for readability I would package them in a dataclass e.g.

@dataclasses.dataclass(slots=True, kw_only=True, frozen=True)
class MdeClusterResult:
    target_possible: float
    pct_change_possible: float
    deff: float
    effective_n: int

src/xngin/stats/cluster_power.py
```diff
@@ -227,24 +229,53 @@ def solve_for_mde_cluster(
...
+
+    clusters_per_arm_list: list[int] = []
+    n_per_arm_list: list[int] = []
+    for prob in arm_probs:
+        n_this_arm = desired_n * prob
+        clusters_this_arm = math.ceil(n_this_arm / metric.avg_cluster_size)
+        clusters_per_arm_list.append(clusters_this_arm)
+        n_per_arm_list.append(round(n_this_arm))
```
qixotic, line 274: Aside: this may not always sum up to desired_n; for assignment.py, we use the "largest remainder method" to allocate any remaining in a not-perfectly-divisible split in the simple random assignment case. If n_per_arm_list is just an estimation reported back to the user for planning, I think it's ok to use this shortcut; assignment.py will allocate entire clusters for the real random assignment.

src/xngin/stats/test_cluster_power.py
```diff
@@ -348,6 +349,65 @@ def test_solve_for_mde_cluster_impl_with_cv():
...
+    assert analysis.effective_sample_size == calculate_effective_sample_size(600, expected_deff)
+
+    # Equal split across 2 arms: 300 each, rounded up to whole clusters of size 30 → 10 per arm.
+    assert analysis.clusters_per_arm == [10, 10]
+    assert analysis.n_per_arm == [300, 300]
+    assert analysis.num_clusters_total == 20
+    assert sum(analysis.clusters_per_arm) == analysis.num_clusters_total
+
+    # The user-facing message should mention the cluster count.
+    assert analysis.msg is not None
```
qixotic, line 368: Please also check analysis.msg.type is as expected

src/xngin/stats/test_cluster_power.py
```diff
@@ -348,6 +349,65 @@ def test_solve_for_mde_cluster_impl_with_cv():
...
+        metric_baseline=100,
+        metric_stddev=20,
+        icc=0.15,
+        avg_cluster_size=30,
+        cv=0.0,
+    )
+
+    analysis = solve_for_mde_cluster(metric=metric, desired_n=600, n_arms=2)
+
+    # MDE math itself is unchanged from the existing _impl tests.
```
qixotic, line 350: This test case looks like a better test of test_solve_for_mde_cluster_impl_basic by calling the wrapper fn directly, so how about just removing the redundant test_solve_for_mde_cluster_impl_basic? The _impl function is pretty simplistic, so testing the wrapper is better.

src/xngin/stats/test_cluster_power.py
```diff
@@ -348,6 +349,65 @@ def test_solve_for_mde_cluster_impl_with_cv():
...
+    )
+
+    # 3:1 weighting on 800 participants → 600 in arm 0, 200 in arm 1.
+    analysis = solve_for_mde_cluster(metric=metric, desired_n=800, n_arms=2, arm_weights=[3.0, 1.0])
+
+    assert analysis.target_n == 800
+    assert analysis.n_per_arm == [600, 200]
+    # 600 / 30 = 20 clusters, 200 / 30 = 6.67 → ceil to 7.
+    assert analysis.clusters_per_arm == [20, 7]
+    assert analysis.num_clusters_total == 27
```
qixotic, line 403: Please also assert the other key values we expect in the response for consistency

qixotic (changes requested):
Mostly refactoring and clarification comments. Happy to merge after the changes, thanks!

src/xngin/stats/cluster_power.py
mdulieu: Done — extracted _normalize_arm_weights at the top of cluster_power.py, used at both call sites.

src/xngin/stats/cluster_power.py
mdulieu: Done — solve_for_mde_cluster_impl now returns MdeClusterResult (frozen, slots, kw_only). solve_for_mde_cluster reads deff and effective_n from the result instead of recomputing. solve_for_sample_size_cluster ignores those two and just pulls target_possible / pct_change_possible.

src/xngin/stats/cluster_power.py
mdulieu: ok

src/xngin/stats/test_cluster_power.py
mdulieu: Done — added assert analysis.msg.type == MetricPowerAnalysisMessageType.SUFFICIENT.

src/xngin/stats/test_cluster_power.py
mdulieu: Done — deleted. Coverage now lives in test_solve_for_mde_cluster_populates_all_fields.

src/xngin/stats/test_cluster_power.py
mdulieu: Done — expanded test_solve_for_mde_cluster_respects_arm_weights to assert target_possible, pct_change_possible, design_effect, effective_sample_size, and the message shape.

mdulieu:
Also flagged while refactoring: solve_for_mde_cluster_impl declared arm_weights but never forwarded it to solve_for_mde_individual_impl, so unbalanced cluster MDE designs silently used balanced-arm math. Fixed (one keyword argument) and covered by a new test_solve_for_mde_cluster_unbalanced_has_larger_mde test that asserts unbalanced splits yield larger MDE than balanced ones. Happy to break this fix out into a separate PR if you'd prefer.

qixotic (approved):
LGTM

PR #254, opened 2026-05-20
User management APIs.

dctaf:
Expands the organization and user management capabilities of the Admin API.

These are the most notable functionality changes:

* Users can create their own orgs 
* Privileged users can manage individual users

This PR implements the backend that will support a series of frontend
improvements.


src/xngin/apiserver/routers/admin/admin_api_types.py
```diff
@@ -121,12 +121,104 @@ class DatasourceSummary(AdminApiBaseModel):
...
+        Field(description="True if the user has logged in at least once."),
+    ]
+    created_at: Annotated[datetime, Field(description="The time the user was created or invited.")]
+
+
+class UserDetail(_UserResponseBase):
+    organizations: Annotated[
+        list[OrganizationSummary],
+        Field(description="Organizations this user is a member of."),
+    ]
```
poornimaramesh, line 157: are we intending to allow non-privileged users to have access to multiple organizations? Just wondering if this is a behaviour we expect to see with our clients?

src/xngin/apiserver/routers/admin/admin_api_types.py
```diff
@@ -121,12 +121,104 @@ class DatasourceSummary(AdminApiBaseModel):
...
+    email: Annotated[str, Field(max_length=MAX_LENGTH_OF_EMAIL_VALUE, description="The user's email address.")]
+
+
+class CreateUserResponse(AdminApiBaseModel):
+    id: Annotated[str, Field(max_length=MAX_LENGTH_OF_ID_VALUE, description="The user's unique ID.")]
+
+
+class OrganizationListItem(AdminApiBaseModel):
+    id: Annotated[str, Field(max_length=MAX_LENGTH_OF_ID_VALUE, description="The unique ID of the organization.")]
+    name: Annotated[str, Field(max_length=MAX_LENGTH_OF_NAME_VALUE, description="The organization's name.")]
```
poornimaramesh, line 190: nit: this class could possibly just inherit `id` and `name` from OrganizationSummary, instead of redefining them here

src/xngin/apiserver/routers/admin/admin_api.py
```diff
@@ -217,6 +225,32 @@ def cache_is_fresh(updated: datetime | None):
...
+    """Raises 403 if no privileged users remain after excluding the given user_id.
+
+    Use this guard before revoking privilege from, or deleting, a user that is privileged.
+    """
+    remaining = await session.scalar(
+        select(func.count())
+        .select_from(tables.User)
+        .where(tables.User.is_privileged.is_(True), tables.User.id != exclude_user_id)
+    )
+    if not remaining:
```
poornimaramesh, line 259: wouldn't `remaining` always be an integer? I suppose better to guard agianst both None and 0 in the same check, just wondering if we would want to make the check explicit?

src/xngin/apiserver/routers/admin/admin_api.py
```diff
@@ -217,6 +225,32 @@ def cache_is_fresh(updated: datetime | None):
     return updated is not None and datetime.now(UTC) - updated < timedelta(minutes=5)
 
 
+def require_privileged(user: tables.User) -> None:
```
poornimaramesh, line 232: since we only use this function within `admin_api.oy`, wouldn't it be neater to chain Dependency injection? i.e. we do `def required_priveleged_user(user: Annotated[tables.User,  Depends(require_user_from_token)])`, and all endpoints that require a priveleged user (e.g. `create_user` on L448 inject `require_privileged as a Dependency`)

src/xngin/apiserver/routers/admin/admin_api.py
```diff
@@ -408,6 +442,227 @@ async def logout(
...
+    body: Annotated[CreateUserRequest, Body(...)],
+) -> CreateUserResponse:
+    """Creates a User record by email. Privileged users only.
+
+    Idempotent: if a user with the given email already exists, the existing user's id is returned
+    and nothing else is changed. Newly-created users have `is_privileged=false` and no organization
+    memberships. When they next sign in via OIDC, the existing user record is bound to their OIDC
+    identity automatically.
+    """
+    require_privileged(user)
```
poornimaramesh: referencing the comment about Dependency injection for priveleged users above

src/xngin/apiserver/routers/admin/admin_api_types.py
```diff
@@ -121,12 +121,104 @@ class DatasourceSummary(AdminApiBaseModel):
...
+            description="Organizations this user is a member of, each with summary counts. Sorted by name ascending."
+        ),
+    ]
+
+
+class PatchUserRequest(AdminApiBaseModel):
+    is_privileged: Annotated[
+        bool | None,
+        Field(description="When non-null, sets the user's privileged status. Privileged status is a system-wide flag."),
+    ] = None
```
poornimaramesh, line 177: nit: I am wondering why allow this to be None? Seems we use the patch request only for the patch endpoint on admin API, wouldn't we be hitting if and only if we wanted to update the status?

src/xngin/apiserver/routers/admin/admin_api.py
```diff
@@ -408,6 +442,227 @@ async def logout(
...
+
+    Cascades to remove all organization memberships. Rejects deleting yourself, and rejects deleting
+    the last privileged user in the system.
+    """
+    require_privileged(user)
+
+    if user_id == user.id:
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail="You cannot delete your own user.",
```
poornimaramesh: yourself?

poornimaramesh (approved):
Works, tests pass, approving. Thanks!

src/xngin/apiserver/routers/admin/admin_api_types.py
dctaf: Yes, this PR intends to do that, but I don't have any expectations around whether users will engage with the ability to have multiple orgs. I anticipate it will be useful for orgs that are mature enough to have test vs prod environments with different membership sets.

src/xngin/apiserver/routers/admin/admin_api_types.py
dctaf: I'm open to this but will defer: these two response types are used for different purposes and may diverge further.

src/xngin/apiserver/routers/admin/admin_api.py
dctaf: Hmm. This is an interesting style + typing concern! I'll leave it as-is for now because we have precedent for this syntax with func.count() queries.

src/xngin/apiserver/routers/admin/admin_api.py
dctaf: Yes! Done. There are other dependencies we can clean up in the APIs too!

src/xngin/apiserver/routers/admin/admin_api.py
dctaf: Done.

src/xngin/apiserver/routers/admin/admin_api_types.py
dctaf: We have precedent (UpdateArmRequest, UpdateDatasourceRequest) for the PATCH endpoints to use None to mean "don't touch this field". We'll probably have more mutable attributes on User in the future, so this follows that pattern. This pattern will break when any of the fields are lists or objects.

src/xngin/apiserver/routers/admin/admin_api.py
dctaf: Done.

PR #258, opened 2026-05-21
Enable PLR0915 and minor refactor to create-testing-dwh.

dctaf:
1. Enables Ruff
[PLR0915](https://docs.astral.sh/ruff/rules/too-many-statements/) and sets a
threshold such that it does not trigger on existing production code. 

2. Refactors `create_testing_dwh` to avoid triggering PLR0915, and adds some minor cleanups for speed/correctness.

### DIFF
--- a/src/xngin/cli/commands/create_testing_dwh.py
+++ b/src/xngin/cli/commands/create_testing_dwh.py
@@ -264,12 +264,8 @@ def _load_bigquery(source: _Source, dest: _Dest) -> None:
     pandas_gbq.to_gbq(df, destination_table, project_id=dest.url.host, if_exists="replace")
 
 
-def _load_postgres(source: _Source, dest: _Dest, *, create_db: bool) -> None:
-    engine = (
-        create_engine_and_database(dest.url)
-        if create_db
-        else create_engine(dest.url, logging_name=SA_LOGGER_NAME_FOR_CLI)
-    )
+def _load_postgres(source: _Source, dest: _Dest) -> None:
+    engine = create_engine(dest.url, logging_name=SA_LOGGER_NAME_FOR_CLI)
     ddl = _get_ddl_magic(source, dest, engine.dialect.identifier_preparer, "postgres")
     with engine.begin() as conn:
         cursor = conn.connection.cursor()
@@ -395,7 +391,7 @@ def create_testing_dwh(
     elif dwh_utils.is_bigquery(dest.url):
         _load_bigquery(source, dest)
     elif dwh_utils.is_postgres(dest.url):
-        _load_postgres(source, dest, create_db=create_db)
+        _load_postgres(source, dest)
     else:
         _err_console.print("Unrecognized database driver.")
         raise typer.Exit(2)
--- a/src/xngin/cli/commands/create_testing_dwh.py
+++ b/src/xngin/cli/commands/create_testing_dwh.py
@@ -184,7 +184,7 @@ def _maybe_create_views(cur, dest: _Dest) -> None:
 
 
 def _table_is_populated(dest: _Dest, *, create_db: bool) -> bool:
-    """Returns True iff the table already exists and contains more one or more rows."""
+    """Returns True iff the table already exists and contains one or more rows."""
     if create_db:
         engine = create_engine_and_database(dest.url)
     else:
--- a/src/xngin/cli/commands/create_testing_dwh.py
+++ b/src/xngin/cli/commands/create_testing_dwh.py
@@ -1,0 +1,412 @@
+"""`create-testing-dwh` command: loads the testing CSV into a dev data warehouse."""
+
+import logging
+import re
+import uuid
+from compression import zstd
+from dataclasses import dataclass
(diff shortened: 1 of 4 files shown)


src/xngin/cli/common.py
```diff
@@ -0,0 +1,56 @@
...
+            url.set(database="postgres"), connect_args=connect_args, logging_name=SA_LOGGER_NAME_FOR_CLI
+        )
+        dwh_utils.extra_engine_setup(engine)
+        with engine.connect().execution_options(isolation_level="AUTOCOMMIT") as conn:
+            # Creating a database can fail in many ways.
+            try:
+                conn.execute(sqlalchemy.text(f"CREATE DATABASE {url.database}"))
+            except psycopg.errors.DuplicateDatabase, psycopg2.errors.DuplicateDatabase:
+                pass
+            except psycopg.errors.IntegrityError, psycopg2.errors.IntegrityError:
```
qixotic, line 46: Thanks for allowing specific errors that are not really a problem for how we use it. But curious, what situation raises an IntegrityError during this `create database` such that we can ignore it? Ah, google tells me that it's most likely due to you having a script running parallel attempts at executing this.

src/xngin/cli/main.py
```diff
@@ -57,95 +46,6 @@ class Base64OrJson(StrEnum):
...
-
-def validate_create_testing_dwh_src(v: Path):
-    allowed_extensions = (".csv", ".csv.zst")
-    for ext in allowed_extensions:
-        if str(v).endswith(ext):
-            return v
-    raise typer.BadParameter("--src must end in .csv or .csv.zst")
-
-
 def parse_iso_datetime(value: str | None) -> datetime | None:
```
qixotic, line 49: why not also move this to common.py?

src/xngin/cli/commands/create_testing_dwh.py
```diff
@@ -0,0 +1,412 @@
...
+    if create_db:
+        engine = create_engine_and_database(dest.url)
+    else:
+        engine = create_engine(dest.url, logging_name=SA_LOGGER_NAME_FOR_CLI)
+    conn = engine.raw_connection()
+    skip = False
+    try:
+        with conn.cursor() as cur:
+            try:
+                _exists(cur, dest)
```
qixotic: isn't _exists() unnecessary if you're going to call _count() anyway? Both would raise if the table never existed, and you always want the count if it does exist.

src/xngin/cli/commands/create_testing_dwh.py
```diff
@@ -0,0 +1,412 @@
...
+
+    df = _read_csv(source)
+    destination_table = f"{dest.url.database}.{dest.table_name}"
+    print("Loading using an inferred schema (warning: may lose fidelity!)...")
+    pandas_gbq.to_gbq(df, destination_table, project_id=dest.url.host, if_exists="replace")
+
+
+def _load_postgres(source: _Source, dest: _Dest, *, create_db: bool) -> None:
+    engine = (
+        create_engine_and_database(dest.url)
```
qixotic: This create should be redundant; looks like it's handled by _table_exists_and_skip() which is always called first, no?

qixotic (approved):
I think there's some redundancy cleanup todo before merge, but otherwise 🆒

src/xngin/cli/common.py
dctaf: Yes! This exception handling change originated a short-lived and abandoned experiment in parallelizing the DWH loads but I kept it for correctness.

src/xngin/cli/commands/create_testing_dwh.py
dctaf: Restored.

src/xngin/cli/main.py
dctaf: parse_iso_datetime is only used in main.py.

src/xngin/cli/commands/create_testing_dwh.py
dctaf: Good catch! Simplified _load_postgres.

src/xngin/cli/commands/create_testing_dwh.py
```diff
@@ -0,0 +1,401 @@
...
+    print(f"Creating views: {', '.join(sorted(view_names))}...")
+    for view_name in view_names:
+        qualified_view_name = f"{dest.schema_name}.{view_name}" if dest.schema_name else view_name
+        cur.execute(
+            f"CREATE OR REPLACE VIEW {qualified_view_name} AS SELECT * FROM {dest.full_table_name}"  # noqa: S608
+        )
+
+
+def _table_is_populated(dest: _Dest, *, create_db: bool) -> bool:
+    """Returns True iff the table already exists and contains more one or more rows."""
```
qixotic:
```suggestion
    """Returns True iff the table already exists and contains one or more rows."""
```

dctaf: Done.

PR #260, opened 2026-05-22
Remove await/async where it is not necessary (and enable RUF029).

dctaf:
Enables RUF029. Removes async/await syntax from methods that don't await. Rename a method for clarity. Allow exceptions to RUF029 for FastAPI touchpoints that would unnecessarily invoke thread pool.


qixotic (approved):
nice

PR #261, opened 2026-05-22
Enables Ruff PT006.

dctaf:
Enables Ruff's [PT006](https://docs.astral.sh/ruff/rules/pytest-parametrize-names-wrong-type/). The primary benefit is more consistent in-IDE type checking and code consistency.


qixotic (approved):
👏

PR #265, opened 2026-05-23
Allow users to re-trigger outbound webhooks.

dctaf:
Allows users to re-trigger an outbound HTTP webhook, strips webhook-tokens from UI, and adds a structured status icon field to the user-facing EventSummary.

Corresponding FE change to enable the feature: https://github.com/agency-fund/evidential-fe/pull/254

### DIFF
--- a/src/xngin/apiserver/routers/admin/admin_api.py
+++ b/src/xngin/apiserver/routers/admin/admin_api.py
@@ -143,7 +143,9 @@ from xngin.apiserver.settings import (
 from xngin.apiserver.snapshots import snapshotter
 from xngin.apiserver.sqla import tables
 from xngin.apiserver.storage.storage_format_converters import ExperimentStorageConverter
+from xngin.events.webhook_sent import WebhookSentEvent
 from xngin.stats import check_power
+from xngin.tq.task_payload_types import WEBHOOK_OUTBOUND_TASK_TYPE
 
 GENERIC_SUCCESS = Response(status_code=status.HTTP_204_NO_CONTENT)
 RESPONSE_CACHE_MAX_AGE_SECONDS = timedelta(minutes=15).seconds
@@ -783,6 +785,35 @@ def convert_events_to_eventsummaries(events):
             )
         )
     return event_summaries
+
+
+@router.post(
+    "/organizations/{organization_id}/events/{event_id}/resend",
+    status_code=status.HTTP_204_NO_CONTENT,
+)
+async def resend_organization_event(
+    organization_id: str,
+    event_id: str,
+    session: Annotated[AsyncSession, Depends(xngin_db_session)],
+    user: Annotated[tables.User, Depends(require_user_from_token)],
+):
+    """Re-enqueues the outbound webhook task that produced a webhook.sent event."""
+    org = await get_organization_or_raise(session, user, organization_id)
+    event = await session.get(tables.Event, event_id)
+    if event is None or event.organization_id != org.id:
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="Event not found.")
+    data = event.get_data()
+    if not isinstance(data, WebhookSentEvent):
+        # Only webhook.sent events can be resent.
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="Event type cannot be resent.")
+    session.add(
+        tables.Task(
+            task_type=WEBHOOK_OUTBOUND_TASK_TYPE,
+            payload=data.request.model_dump(),
+        )
+    )
+    await session.commit()
+    return GENERIC_SUCCESS
 
 
 @router.post("/organizations/{organization_id}/members", status_code=status.HTTP_204_NO_CONTENT)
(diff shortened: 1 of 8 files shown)


src/xngin/tq/task_payload_types.py
qixotic, line 17: As an aside, what do you think about typing the body here with a discriminated type called `WebhookBody` in webhook_types.py, that collects the union of all body types there, and have each type define its own ui_link() helper? Then the `WebhookSentEvent` can define its `BaseEventModel::link()` that delegates to the body's helper. So in `ExperimentCreatedWebhookBody`'s case, it would create the admin-api friendly `"/datasources/{body.datasource_id}/experiments/{body.experiment_id}"` relative url. 

This would also let us include the `WebhookBody.type` as part of the as part of `WebhookSentEvent::summarize()` to make it clearer what subtype of webhook event was sent, e.g. "Sent POST for experiment.created to http://localhost:8000 (failed)"

src/xngin/events/webhook_sent.py
qixotic, line 19: Could we prefix the summary with an ✅/❌ emoji and the (success/failed) text instead of suffixing it? Think it would make it easier to spot at a glance.

src/xngin/apiserver/routers/admin/admin_api.py
```diff
@@ -779,12 +781,41 @@ def convert_events_to_eventsummaries(events):
...
+    """Re-enqueues the outbound webhook task that produced a webhook.sent event."""
+    org = await get_organization_or_raise(session, user, organization_id)
+    event = await session.get(tables.Event, event_id)
+    if event is None or event.organization_id != org.id:
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="Event not found.")
+    data = event.get_data()
+    if not isinstance(data, WebhookSentEvent):
+        # Only webhook.sent events can be resent.
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="Event type cannot be resent.")
+    session.add(
```
qixotic, line 810: One thing I noticed was that if I resend an event that already has a pending task for it, this enqueues a 2nd task without touching the first. So if the first hasn't already been marked 'success' or 'dead', we could have 2 tasks firing for the same event, regardless of the status of the other.

What if we added an optional many:1 relation from an Event to a Task when the Event is created? Then we could look for any non-final-state task associated with the event, and if present mark it as 'dead' or a new state 'resend' in the same transaction as this session.add()? The very first Task for a new experiment though wouldn't be associated with a WebhookSentEvent.TYPE, but rather tied to the ExperimentCreatedEvent.TYPE as part of `commit_experiment_impl()`. This also means `webhook_outbound_handler()` would need to also pass the task id to `_record_webhook_sent_event()`.

But would we ever want multiple Tasks to spawn from a single Event?

qixotic (approved):
As a quicker fix, this looks good, particularly with the FE dialog warning of being able to receive multiple events for the same exp, thanks!

I've left some suggestions as well for possible UX improvements around webhook events to consider here or in future PRs.

src/xngin/apiserver/routers/admin/admin_api.py
```diff
@@ -779,12 +781,41 @@ def convert_events_to_eventsummaries(events):
...
+):
+    """Re-enqueues the outbound webhook task that produced a webhook.sent event."""
+    org = await get_organization_or_raise(session, user, organization_id)
+    event = await session.get(tables.Event, event_id)
+    if event is None or event.organization_id != org.id:
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="Event not found.")
+    data = event.get_data()
+    if not isinstance(data, WebhookSentEvent):
+        # Only webhook.sent events can be resent.
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="Event type cannot be resent.")
```
abhishek-compro, line 809: All these can be a `Depends` which would reduce lot of boiler plate required per route. Its okay if this is only being used here but do give it a thought if this is being repeated at multiple places.

abhishek-compro (changes requested):

src/xngin/events/webhook_sent.py
qixotic: This was the one tweak that I would like to see folded into this PR @dctaf . Thanks!

src/xngin/tq/task_payload_types.py
dctaf: We should definitely consider usability improvements like this, but I'd like to do so as part of a larger reconsideration of how we present events and webhook statuses generally.

src/xngin/events/webhook_sent.py
dctaf: I've addressed this by synthesizing a `status_icon` field.

src/xngin/apiserver/routers/admin/admin_api.py
dctaf: I agree, this would be a nice cleanup! Some of the endpoints have intentionally avoided using dependencies for entity reads to minimize database round trips on performance-sensitive endpoints (e.g. get_assignment for online experiments), but most of them can be cleaned up with reusable dependencies.

PR #267, opened 2026-05-25
Step 1: MAB DWH target binding

markbotterill:
## Ticket

[Here](https://github.com/orgs/agency-fund/projects/8?pane=issue&itemId=180935203&issue=agency-fund%7Cevidential-sprint%7C306)

## Related PRs

None yet - but reference design docs for this specific phase is [here](https://www.notion.so/agencyfund/1-Ready-for-review-Backend-New-DWH-validated-MAB-targets-3626311d558d8063b4fffea16af2d836?source=copy_link)

## Description

This is the first step of the DWH-backed MAB journey. It involves:
1. Introducing a new type of experiment in the backend - `MABDwhExperiment`
2. Add the ability for this experiment to read an external data source
3. Add the ability to save a selected column as the "target" column for what we're trying to check for
4. Add the ability to validate an incoming result at POST time. 

## Future Tasks (optional)

Remaining tasks on the roadmap [here](https://www.notion.so/agencyfund/DWH-Connector-for-C-MABs-3606311d558d801a97b7ca328bdeeb49?source=copy_link)

## How has this been tested?

New tests added and pass!

## Checklist

Fill with `x` for completed.


src/xngin/apiserver/routers/admin/admin_api.py
```diff
@@ -1593,15 +1594,15 @@ async def analyze_experiment(
...
+        case MABExperimentSpec() | MABDwhExperimentSpec():
             return await experiments_common.analyze_experiment_bandit_impl(xngin_session, experiment)
         case CMABExperimentSpec():
             raise LateValidationError(
                 """Invalid experiment type for bandit analysis; for CMAB experiments,
                 use the corresponding POST endpoint.""",
             )
         case _:
-            assert_never()
+            assert_never(design_spec)
```
dctaf, line 1946: 👍

src/xngin/apiserver/routers/experiments/experiments_common.py
```diff
@@ -144,6 +146,49 @@ async def fetch_fields_or_raise(
...
+        )
+
+    target_type = referenced_fields_and_types[design_spec.target_field_name]
+    if not target_type.is_supported_as_metric():
+        raise LateValidationError(
+            f"Invalid target field '{design_spec.target_field_name}' of type {target_type.value}. "
+            "Only boolean or numeric data types are supported as targets."
+        )
+
+    return referenced_fields_and_types
```
dctaf, line 200: This block shares a lot with convert_table_to_fields_or_raise. Is it possible to reuse the common parts? Then the mab and non-mab implementations might both be fewer lines of code, and perhaps we'll get to re-use error handling for the two LateValidationError cases instead of duplicating them..

e.g. something like this:
```python
async def fetch_dwh_fields_or_raise(...):
  async with DwhSession(datasource.get_config().dwh) as dwh:
    sa_table = await dwh.inspect_table(design_spec.table_name)
  return convert_table_to_fields_or_raise(sa_table, 
    design_spec,
    metric_fields=..., 
    additional_fields={**filter_fields, **strata_fields})
  return ref_fields_and_types

async def fetch_mab_dwh_fields_or_raise(...):
  async with DwhSession(datasource.get_config().dwh) as dwh:
    sa_table = await dwh.inspect_table(design_spec.table_name)
  return convert_table_to_mab_dwh_fields_or_raise(sa_table, 
    design_spec, 
    metric_fields=[design_spec.target_field_name], 
    additional_fields=[design_spec.primary_key])
```

src/xngin/apiserver/routers/common_api_types.py
```diff
@@ -1092,6 +1092,40 @@ class MABExperimentSpec(BaseBanditExperimentSpec):
     experiment_type: Literal[ExperimentsType.MAB_ONLINE] = ExperimentsType.MAB_ONLINE
 
 
+class MABDwhExperimentSpec(BaseBanditExperimentSpec):
```
dctaf, line 1084: I recommend a model validator to enforce that primary_key != target_field_name. This is a useful invariant to catch early, and also, without this, it looks like the code that creates ExperimentFields will try to create duplicate rows for the same field. See check_has_only_one_of_pct_change_or_target for an example.

dctaf (changes requested):
This is a brief first pass review, but this is looking good!!

src/xngin/apiserver/routers/experiments/experiments_common.py
```diff
@@ -592,7 +650,7 @@ async def get_existing_assignment_for_participant(
                     tables.ArmAssignment.participant_id == participant_id,
                 )
             )
-        case ExperimentsType.MAB_ONLINE | ExperimentsType.CMAB_ONLINE:
+        case ExperimentsType.MAB_ONLINE | ExperimentsType.MAB_ONLINE_DWH | ExperimentsType.CMAB_ONLINE:
```
qixotic, line 660: This might be a good time to introduce a factory method in ExperimentAnalysisType to map an ExperimentsType to one of FREQ or BANDIT. Maybe even rename ExperimentAnalysisType to ExperimentFamily? Then for situations like this and elsewhere, we could just `match ExperimentFamily.from_type(experiment_type):`.

src/xngin/apiserver/storage/storage_format_converters.py
```diff
@@ -41,13 +41,87 @@ def _set_power_response_json(experiment: tables.Experiment, value: capi.PowerRes
         experiment.power_analyses = capi.PowerResponse.model_validate(value).model_dump()
 
 
+def _apply_design_spec_filters(
```
qixotic: Was this extracted in anticipation of reuse in the future by the new MABDwhExperimentSpec?

I've merged an alternate refactoring in #271, in particular with the use of an `_upsert_field_used` helper to reduce a lot of the code duplication I had originally around incrementally creating new tables.ExperimentField objects or updating them based on the fields used by the design spec. It ensures that a table's field name used in the spec only ever has one corresponding tables.ExperimentField entry even if it shows up in a few use cases (like both a Filter and as Strata). It also groups that frequentist spec logic into its own `_make_freq_experiment_fields` helper, analogous to your `_set_mab_dwh_experiment_fields` below.

src/xngin/apiserver/storage/storage_format_converters.py
```diff
@@ -516,6 +567,9 @@ def init_from_components(
                     for i, arm in enumerate(design_spec.arms, start=1)
                 ]
 
+                if isinstance(design_spec, capi.MABDwhExperimentSpec):
```
qixotic: Combine this block with your earlier one that sets datasource_table?

qixotic (commented):
A few more to add to Doug's review. Looking forward to this big step!

src/xngin/apiserver/routers/common_api_types.py
markbotterill: Great point - implemented!

src/xngin/apiserver/routers/experiments/experiments_common.py
markbotterill: Nice catch - sorted!

src/xngin/apiserver/routers/experiments/experiments_common.py
markbotterill: I do like this idea but think it might increase scope of this PR to be a bit large and overreaching! It is getting kinda clunky to maintain all of these `|`s so I can raise a separate PR for it 🫡

markbotterill: Issue raised [here](https://github.com/agency-fund/evidential-sprint/issues/359) for it!

src/xngin/apiserver/snapshots/test_snapshotter.py
```diff
@@ -57,7 +58,7 @@ async def make_experiment(
     match design_spec:
         case PreassignedFrequentistExperimentSpec() | OnlineFrequentistExperimentSpec():
             field_type_map = await fetch_fields_or_raise(datasource, design_spec)
-        case MABExperimentSpec() | CMABExperimentSpec():
+        case MABExperimentSpec() | MABDwhExperimentSpec() | CMABExperimentSpec():
```
qixotic: shouldn't the MABDwhExperimentSpec() case call fetch_mab_dwh_fields_or_raise?

src/xngin/apiserver/routers/common_api_types.py
```diff
@@ -1081,6 +1081,40 @@ class MABExperimentSpec(BaseBanditExperimentSpec):
...
+    target_field_name: Annotated[
+        FieldName,
+        Field(
+            description=(
+                "Column name in table_name whose values this bandit optimises. "
+                "Its data type is resolved from the DWH at create-time and used to validate "
+                "subsequent outcome reports."
+            ),
+        ),
+    ]
```
qixotic, line 1115: Don't see the model validator checking that primary_key != target_field_name as dctaf suggested earlier.

src/xngin/apiserver/routers/experiments/experiments_common.py
```diff
@@ -144,6 +146,49 @@ async def fetch_fields_or_raise(
         return convert_table_to_fields_or_raise(sa_table, design_spec)
 
 
+async def fetch_mab_dwh_fields_or_raise(
```
qixotic, line 149: You kept this independent just for the MABDwh support; was that intentional? Your reply to @dctaf suggested you were looking to share logic.

I think rather than Doug's original suggestion, if you wanted to share the logic better with `convert_table_to_fields_or_raise`, I think we'd have to wrap the `referenced_fields` and `metric_fields` construction into a case-match section (or via helper fn, as the freq cases are more involved). The LateValidationError then becomes a little more generic for the mabdwh case (no reporting of target_type.value). 

And lastly there could be case-specific validation for experiment types after the general metric validation, as that's used for the frequentist path still, i.e. the `    for filter_ in design_spec.filters:` block. 

This means there could be up to two more `match` sections though, which may be slightly more annoying to read.

But in doing so we can use `fetch_fields_or_raise()` as a common call for all the dwh types, while the power_check() API can still call the helper `convert_table_to_fields_or_raise()` for field validation, even with the shared logic.

src/xngin/apiserver/routers/experiments/test_experiments_common.py
```diff
@@ -1485,6 +1536,87 @@ async def test_create_experiment_impl_for_cmab_online(xngin_session, testing_dat
...
+    ("target_field_name", "expected_data_type"),
+    [
+        ("is_onboarded", "boolean"),
+        ("current_income", "numeric"),
+    ],
+)
+async def test_create_experiment_impl_for_mab_dwh_online(
+    xngin_session, testing_datasource, target_field_name: str, expected_data_type: str
+):
+    """MAB-DWH happy path: target column type is resolved from DWH and persisted on ExperimentField rows."""
```
qixotic, line 1549: Could you add a happy path test to test_admin_api.py as well that just uses the APIs via the respective clients (see other tests using aclient+eclient) to both create this new type of experiment as well as update an outcome? Thanks!

qixotic (changes requested):

src/xngin/apiserver/routers/common_api_types.py
markbotterill: Here now! Was not pushed before!

src/xngin/apiserver/snapshots/test_snapshotter.py
markbotterill: Good catch! Updated.

src/xngin/apiserver/routers/experiments/experiments_common.py
qixotic: ah [571ea13](https://github.com/agency-fund/evidential-be/pull/267/commits/571ea1385bb18c02fef68811ebbc682a1880f7d8) looks fine 👍

qixotic (approved):
👍  Please add a brief test of MABDwhExperimentSpec to test_admin_api.py before merging, thanks!

src/xngin/apiserver/routers/experiments/test_experiments_common.py
markbotterill: Done 🫡

markbotterill:
Thank you for reviews @qixotic and @dctaf - finally merging in 🥳

PR #268, opened 2026-05-27
309 better error handling for turnio endpoints

poornimaramesh:
## Ticket

[Issue 309](https://github.com/agency-fund/evidential-sprint/issues/309)

## Description
- Standardized error handling, and surface errors from Turn.io under a single error code on Evidential
- Added tests for this behaviour

## How has this been tested?
Run tests, check endpoints work with curl requests

## Checklist


src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
```diff
@@ -67,15 +67,62 @@
...
+    Any non-2xx response from Turn.io and httpx.RequestErrors (e.g. network issues, timeouts) are logged
+    and re-raised as 502 HTTP exceptions, but with appropriate status codes and error messages
+    reproduced for debugging.
+    """
+    if headers is None:
+        headers = {}
+    headers.update({"Authorization": f"Bearer {turn_api_token}"})
+
+    try:
+        async with httpx.AsyncClient(timeout=TURN_REQUEST_TIMEOUT_SECONDS) as client:
```
dctaf: To make this more robust, you could also add transport-layer retries: https://www.python-httpx.org/advanced/transports/. We also have a dependency that provides a basic retrying transport (https://github.com/agency-fund/evidential-be/blob/main/src/xngin/apiserver/dependencies.py#L21), but up to you whether to re-use that!

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
```diff
@@ -67,15 +67,62 @@
...
+        ) from exc
+    except httpx.HTTPStatusError as exc:
+        logger.error(
+            f"Turn.io API returned non-2xx status at {method} {url}: {exc.response.status_code} - {exc.response.text}"
+        )
+        raise HTTPException(
+            status_code=status.HTTP_502_BAD_GATEWAY,
+            detail=f"Turn.io API returned non-2xx status. Details: {exc.response.status_code} - {exc.response.text}",
+        ) from exc
+    return response
```
dctaf, line 114: This removes the {401, 403} => 409 handling  (useful for the "Please check that the configured API token is valid." guidance, perhaps?), and the contextualized log lines, from the previous implementation. Is that intentional?

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
```diff
@@ -67,15 +67,62 @@
...
     },
 }
 
 
+async def _call_turn_api(
+    turn_api_token: str,
+    method: str,
+    url: str = TURN_JOURNEYS_URL,
+    headers: dict[str, str] | None = None,
+    **request_kwargs: Any,
```
dctaf: Consider removing arguments that are not used.

dctaf (changes requested):

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
poornimaramesh: Yes, this is intentional. The Turn.io API provides informative enough messages in case of API token errors, so it allows us to standardize the error codes we return (and not go down the rabbit hole of handling each error the Turn API might return separately, as we discussed doing [before](https://github.com/agency-fund/evidential-be/pull/232#discussion_r3132449887)).

Would you recommend rolling back these changes?

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
poornimaramesh: using the dependency, thanks for pointing me to it!

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
poornimaramesh: removed, thanks!

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
dctaf: Your expertise on the Turn platform should guide the decision here.  If you're comfortable with any 2xx error codes being considered a successful response, then this will work.

TURN_JOURNEYS_RESPONSES should be updated with the new expected response codes; it currently says that this method returns a 409, but perhaps we now only return 404 and 502.

dctaf (approved):

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
poornimaramesh: Thanks! I haven't seen Turn's Journeys API return a non-200 2xx error code, and from [all documentation](https://whatsapp.turn.io/docs/api/errors) that I can find we get 2xx error codes only on requests that Turn forwards to Meta (which is not the case here).
I'll merge as is now (I've updated the expected response codes), but I'll confirm with Simon / Turn.io team at our next monthly check in.

Issue #217, opened 2026-04-10
[Frontend] Extend UI to support cluster randomization


mdulieu:
## Background

This issue tracks the frontend UI for cluster randomization power analysis, as called out in #154 (Phase 3). The backend is complete across PRs #159, #160, and #161.

> **Note on terminology:** the existing "MME" label throughout the frontend should be renamed to **"Target MDE"** as part of this work (consistent with #216). This affects the Targeting and Design modal and the Experiment Details analysis bar.

### Parameters page — Randomization section

The Parameters page has two relevant sections:

**Metrics section:** a dropdown to select the metric, with a Minimum Effect (% change) field.

**Randomization section:** a **Strata / Cluster** toggle. The user must choose one:
- **Strata tab:** shows a dropdown to select strata fields for balanced randomization.
- **Cluster tab:** shows a **Cluster ID field** dropdown. Once a field is selected, three fields are automatically calculated from the datasource — **ICC**, **CV**, and **Avg cluster size** — each overridable via a pencil icon. If CV > 0.5, a **⚠ high variability** warning appears below the CV field.

**ClustersUI1A** — Strata tab (default state)

<img width="721" height="699" alt="Image" src="https://github.com/user-attachments/assets/ad3e92fa-2205-4535-8151-a19e08d67f35" />

**ClustersUI1B** — Cluster tab selected, no field chosen yet

<img width="715" height="693" alt="Image" src="https://github.com/user-attachments/assets/0edc1457-b960-46a3-ba31-b4dd4f6a235e" />

**ClustersUI1C** — Cluster tab selected, field chosen, ICC/CV/avg cluster size populated

<img width="726" height="772" alt="Image" src="https://github.com/user-attachments/assets/a54241eb-9618-4b65-9cac-785d112187c5" />

---

### Power Analysis page

All quantities are expressed as **clusters first, participants underneath** throughout — in the success message, the metric card (Required / Available / Available non-null), and the Select Target Sample Size options.

If CV > 0.5, a **⚠ high variability** warning appears inside the result box.

In the **Use custom sample size** option, there are two linked input fields: **Clusters** and **Participants**. Filling either one triggers automatic recalculation of the other after a 0.5-second debounce.

**ClustersUI2A** — Custom sample size option empty (placeholder state)

<img width="690" height="730" alt="Image" src="https://github.com/user-attachments/assets/bed12a28-fb71-41e2-b83f-767f61352f98" />

[...]

PR #269, opened 2026-05-28
Persist cluster_key with ArmAssignments

qixotic:
## Description

As part of supporting cluster-randomized experiments, this PR stores the assigned unit's cluster_key value with its ArmAssignment in the db, for later use in analysis. 
(Related issue: #217)

## Future Tasks

- Use persisted cluster_key values at analysis time.
- (nice to have) Warn users if the cluster_key values have drifted since the time of random assignment.
- move cluster_key from BaseFrequentistDesignSpec => PreassignedFrequentistExperimentSpec to avoid needing to do extra validation around whether it's set only for the right frequentist type 

## How has this been tested?

unittests


src/xngin/apiserver/routers/common_api_types.py
```diff
@@ -1191,6 +1192,15 @@ class Assignment(ApiBaseModel):
             max_length=MAX_LENGTH_OF_PARTICIPANT_ID_VALUE,
         ),
     ]
+    cluster_key: Annotated[
+        str | None,
+        Field(
+            description=(
+                "Cluster identifier for this participant. Present for cluster-randomized preassigned experiments."
+            ),
+            max_length=MAX_LENGTH_OF_PARTICIPANT_ID_VALUE,
```
dctaf: Consider introducing a separate constant for this maximum length instead of using the same max_length as the participant_id fields.

src/xngin/apiserver/routers/experiments/experiments_common.py
dctaf, line 769: Do we need to consider cluster_key here too?

src/xngin/apiserver/routers/common_api_types.py
```diff
@@ -1191,6 +1192,15 @@ class Assignment(ApiBaseModel):
             max_length=MAX_LENGTH_OF_PARTICIPANT_ID_VALUE,
         ),
     ]
+    cluster_key: Annotated[
```
dctaf, line 1196: Perhaps we should validate that cluster_key is not set for the types of experiments that do not support it (e.g. online frequentist).

dctaf (changes requested):

src/xngin/apiserver/routers/experiments/experiments_common.py
qixotic: That function is misleading; higher up it returns None if it's FREQ_PREASSIGNED.

I've pushed a commit that consolidate the per-type logic into one place within the try block. Although they don't need to be within the try, they also don't raise IntegrityErrors and aren't too many lines, so I think the logical grouping is worth it.

src/xngin/apiserver/routers/common_api_types.py
qixotic: You suggested elsewhere that since we're only supporting preassigned in practice (at least for the foreseeable future) how about making it a direct member of PreassignedFrequentistExperimentSpec. That would also not preclude supporting it one day in the future for OnlineFrequentistExperimentSpec as well if needed. I like that idea, but will defer that to a followup PR. Thanks!

dctaf (approved):

qixotic:
cc @mdulieu

PR #274, opened 2026-06-02
Wire up cluster randomized experiment analysis to API

qixotic:
## Description

Followup to #269: use the persisted cluster_key values in the analysis of a preassigned cluster randomized experiment, by passing those values into `analyze_freq_experiment`.

## Future Tasks

Extend UI to report when we're looking at a cluster-randomized experiment so the user knows it's being analyzed accordingly. Related: #217 

## How has this been tested?

unittests

### DIFF
--- a/src/xngin/apiserver/routers/experiments/experiments_common.py
+++ b/src/xngin/apiserver/routers/experiments/experiments_common.py
@@ -1123,14 +1123,26 @@ async def analyze_experiment_freq_impl(
     )
 
 
-async def read_assignments_efficiently(xngin_session: AsyncSession, experiment_id: str) -> tuple[list[str], DataFrame]:
+async def read_assignments_efficiently(
+    xngin_session: AsyncSession,
+    experiment_id: str,
+    *,
+    include_cluster_key: bool = False,
+) -> tuple[list[str], DataFrame]:
     """Reads assignments directly from Postgres via a COPY statement.
 
     Reads CSV output in row-bounded chunks and concatenates the parsed frames.
     """
-    select_query = t"SELECT arm_id, participant_id FROM arm_assignments WHERE experiment_id = {experiment_id}"  # type: ignore
+    if include_cluster_key:
+        select_query = (
+            t"SELECT arm_id, participant_id, cluster_key FROM arm_assignments WHERE experiment_id = {experiment_id}"  # type: ignore
+        )
+        column_names = ["arm_id", "participant_id", "cluster_key"]
+    else:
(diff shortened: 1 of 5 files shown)


qixotic:
cc @mdulieu

src/xngin/apiserver/routers/experiments/experiments_common.py
```diff
@@ -1032,9 +1032,14 @@ async def analyze_experiment_freq_impl(
...
 
-    participant_ids, assignments_df = await read_assignments_efficiently(xngin_session, experiment.id)
+    include_cluster = experiment.cluster_key_field() is not None
+    participant_ids, assignments_df = await read_assignments_efficiently(
+        xngin_session, experiment.id, include_cluster_key=include_cluster
+    )
     if assignments_df.empty:
         raise StatsAnalysisError("No participants found for experiment.")
+
+    cluster_col = "cluster_key" if include_cluster else None
```
poornimaramesh: it looks like we're re-using the "cluster_key" as a literal in multiple functions. Can we turn it into a global variable somewhere (maybe in `constants.py`) that we reuse everywhere, in case we ever decide to change it?

src/xngin/apiserver/routers/experiments/experiments_common.py
```diff
@@ -1123,22 +1129,34 @@ async def analyze_experiment_freq_impl(
...
     """Reads assignments directly from Postgres via a COPY statement.
 
     Reads CSV output in row-bounded chunks and concatenates the parsed frames.
     """
-    select_query = t"SELECT arm_id, participant_id FROM arm_assignments WHERE experiment_id = {experiment_id}"  # type: ignore
+    if include_cluster_key:
+        select_query = (
+            t"SELECT arm_id, participant_id, cluster_key FROM arm_assignments WHERE experiment_id = {experiment_id}"  # type: ignore
+        )
+        column_names = ["arm_id", "participant_id", "cluster_key"]
```
poornimaramesh: see comment above about turning this into a global variable

poornimaramesh (approved):
Looks good to me, tests pass, approving. Thanks!

src/xngin/apiserver/routers/experiments/experiments_common.py
qixotic: Thanks; that code also wasn't too consistent with the other hard-coded assumptions baked into the original `analyze_experiment` either. I've changed the analysis function to accept column names for both the participant and arm ids as well now (removing the hard-coded assumptions within). 👍

src/xngin/apiserver/routers/experiments/experiments_common.py
qixotic: Explicitly constructed from our tables.py names now. 👍

PR #275, opened 2026-06-03
Bump FastAPI, Pydantic, and fastapi-typed-client-fork.

dctaf:
The primary motivation of this PR is to move us to a fork of
fastapi-typed-client that more easily aligns with the upstream implementation.
They've implemented some (but not all) of my improvements, and now we will be
able to track them more easily.

This also bumps FastAPI & Pydantic. The newer Pydantic adjusts some internals
of how discriminated unions are handled and I've done extensive experiments to
validate that they don't affect us.


poornimaramesh (approved):
Looks good to me, approving, thanks!

PR #276, opened 2026-06-03
CI: Replace pre-commit with prek, and use ubuntu-slim for simple jobs.

dctaf:
- **Conditionally invoke ruff-annotations and move some jobs to ubuntu-slim.**
- **Switch pre-commit job to prek.**
- **Formatting.**


poornimaramesh (approved):
Looks good to me, thanks!

PR #277, opened 2026-06-04
Disallow deletes of resources from unrelated collections.

dctaf:
Fixes a bug where we accidentally allowed subresources to be deleted by
referring to them via unrelated hierarchies; e.g. deletes of /a/A/b/B and
a/A/b/B' would succeed even if B' wasn't in A.


src/xngin/apiserver/routers/admin/test_admin_api.py
```diff
@@ -994,6 +1006,25 @@ def test_delete_datasource(testing_datasource, aclient: AdminAPIClient, aclient_
...
+
+    with expect_status_code(404):
+        aclient_unpriv.delete_datasource(
+            organization_id=caller_org_id,
+            datasource_id=testing_datasource.datasource_id,
+        )
+
+    response = aclient.get_datasource(datasource_id=testing_datasource.datasource_id).data
+    assert response.id == testing_datasource.datasource_id
+
```
qixotic, line 1026: Can you also assert these two cross-org combinations here:

    # aclient is privileged but still needs to be a member of caller_org_id
    with expect_status_code(403):
        aclient.delete_datasource(
            organization_id=caller_org_id,
            datasource_id=_caller_datasource_id,
        )

    # aclient can access datasource_id but not via caller_org_id
    with expect_status_code(403):
        aclient.delete_datasource(
            organization_id=caller_org_id,
            datasource_id=testing_datasource.datasource_id,
        )

(The combination of aclient_unpriv deleting <testing_datasource.org_id, testing_datasource.datasource_id> was already handled in the original `test_delete_datasource`.)

qixotic (approved):
Great catch!

src/xngin/apiserver/routers/admin/test_admin_api.py
dctaf: Done

PR #281, opened 2026-06-08
Journey caching and invalidation

poornimaramesh:
## Ticket

Fixes [Issue 311](https://github.com/agency-fund/evidential-sprint/issues/311)

## Related PRs

Frontend: [PR 263](https://github.com/agency-fund/evidential-fe/pull/263)

## Description
turn APIs

## Future Tasks (optional)
(Will create JIRA ticket for this when we migrate) -- add an endpoint for a Turn App webhook to hit, and trigger a refresh of the Journey list
Making sure that this endpoint works, and is secure (i.e. not vulnerable to malicious attacks is more non-trivial than I thought, so will deal with it in a separate PR).

## How has this been tested?
Run tests, `curl` for endpoints

## Checklist


src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
```diff
@@ -114,6 +112,33 @@ async def _call_turn_api(
...
+    turn_connection: tables.TurnConnection,
+    httpx_client: httpx.AsyncClient,
+    commit_session: bool = True,
+) -> dict[str, str]:
+    """Refreshes the cached Turn.io journeys on the TurnConnection.
+
+    Optionally commits the session after updating the TurnConnection (defaults to True).
+    This can be set to False when this function is called within a larger transaction that
+    should be committed all at once, such as when setting/rotating the API token, since that
+    operation also updates the TurnConnection and we want to avoid committing twice in a row.
```
dctaf: Our existing convention is that the API methods (in the `_api.py` files) are responsible for invoking `.commit()`. That pattern makes the API method lifecycle roughly equivalent to the lifetime of the database transaction, which is easy to reason about. If we allow commits in the middle of a request's work, then we might have API requests that have incomplete side effects.

Coupling database commits to the API handlers will also (eventually) make it easier for us implement the superior SERIALIZABLE isolation level. 

Because this refresh_journeys_dict method is only ever called with commit_session=False, I recommend we simplify the implementation by removing the commit_session argument and making the caller responsible for committing.

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
```diff
@@ -114,6 +112,33 @@ async def _call_turn_api(
...
+    response = await _call_turn_api(
+        httpx_client=httpx_client, turn_api_token=turn_connection.get_turn_api_token(), method="GET"
+    )
+    journeys_dict = {journey["name"]: journey["uuid"] for journey in response.json()}
+    turn_connection.journeys_dict = journeys_dict
+
+    if commit_session:
+        await session.commit()
+
+    return journeys_dict
```
dctaf: Instead of modifying the turn_connection entity, consider simplifying the method signature and letting the caller set the return value on the entity:

```python
def set_organization_turn_connection(...):
  # ...
  turn_connection.journeys_dict = refresh_journeys_dict(httpx_client, turn_connection.get_turn_api_token()).
```

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
```diff
@@ -114,6 +112,33 @@ async def _call_turn_api(
...
+    This can be set to False when this function is called within a larger transaction that
+    should be committed all at once, such as when setting/rotating the API token, since that
+    operation also updates the TurnConnection and we want to avoid committing twice in a row.
+
+    Returns the updated journey list.
+    """
+    response = await _call_turn_api(
+        httpx_client=httpx_client, turn_api_token=turn_connection.get_turn_api_token(), method="GET"
+    )
+    journeys_dict = {journey["name"]: journey["uuid"] for journey in response.json()}
```
dctaf: This will raise KeyError if the items in the JSON array don't have .name or .uuid keys. If the response was parsed by a Pydantic model, you'd get some validation almost for free, and could catch the ValidationError and turn it into a useful error message for the caller.

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
```diff
@@ -364,7 +391,21 @@ async def get_turn_arm_journey_mapping(
...
         )
 
-    return GetTurnArmJourneyMappingResponse(arm_to_journeys=mapping.arm_journey_map)
+    turn_journeys_row = await session.execute(
+        select(tables.TurnConnection.journeys_dict).where(
+            tables.TurnConnection.organization_id == ds.organization_id,
+        )
+    )
+    turn_journeys = turn_journeys_row.scalar_one_or_none()
+    uuids = list(turn_journeys.values()) if turn_journeys else []
```
dctaf: `turn_journeys` is falsey when it is `{}`, but it looks like `refresh_journeys_dict` will accept an empty list of journeys.

Should this be `if turn_journeys is None`? 
Should we make `TurnConnection.turn_journeys` non-nullable?
Is the concept of "staleness" honored when there are zero journeys returned by the Turn API?

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
dctaf: Nit: consider renaming this variable to `turn_config`.

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
```diff
@@ -149,27 +189,23 @@ async def set_organization_turn_connection(
         turn_connection = tables.TurnConnection(organization_id=org.id)
         turn_connection.set_turn_api_token(body.turn_api_token)
         session.add(turn_connection)
-        await session.commit()
 
     elif turn_connection.get_turn_api_token() != body.turn_api_token:
```
dctaf, line 189: If the user sets a new token from an entirely different Turn project, this could potentially invalidate the ExperimentTurnConfig.arm_journey_map entries. I see the intention is to handle these by surfacing the "stale" journeys via get...mapping(). This took me some effort to understand, and at first glance it /looks/ like a bug, so consider adding a comment: `# This could make the arm_journey_ids stored in ExperimentTurnConfig.arm_journey_map invalid. We expose that drift to the UI via get_..mapping()'s reporting of stale journey IDs.`

dctaf (changes requested):
Apologies for the delay!

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
poornimaramesh: I've updated the conditional to explicitly check for nulls.

However, I'd like to allow `refresh_journeys_dict` to return empty journey lists and for `TurnConnection.turn_journeys` to be nullable. This is because I don't think we have enough evidence to be opinionated about when users should be setting up Journeys for an experiment (i.e. before they design the experiment on Evidential, or after). 

For deployments with partners using Turn.io I've seen it go both ways, so I'd like to allow for maximum flexibility now, but I'm hoping a few more deployments will tell us if it makes user flow / onboarding easier if we force them to set up Journeys first.

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
poornimaramesh: Done, thanks for the explanation! I force the caller to commit the session now.

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
poornimaramesh: Done, thanks!

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
poornimaramesh: Added the comment, thanks! And sorry about the confusion :(

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
poornimaramesh: done, thanks!

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
poornimaramesh: added pydantic validation, thanks!

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
```diff
@@ -94,6 +100,8 @@ async def _call_turn_api(
             headers=headers,
         )
         response.raise_for_status()
+        journeys_dict = [Journey.model_validate(journey) for journey in response.json()]
```
dctaf: Nit: journeys_dict is now a list. . When the variable name contains a hint about the type, it's extra confusing if the types don't match. Consider renaming this to `journeys`.

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
```diff
@@ -301,9 +301,9 @@ async def get_organization_turn_journeys(
             "Re-set the Turn.io API token to fetch Journeys from Turn.io and store them.",
         )
 
-    journeys_dict = turn_connection.journeys_dict
+    journeys = [Journey(name=name, uuid=uuid) for name, uuid in turn_connection.journeys_dict.items()]
```
dctaf: Consider adding a sort to this (maybe by uuid) so that the returned order is stable. This can prevent flashing in the UI and lead to more deterministic tests.

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api_types.py
```diff
@@ -27,18 +27,19 @@ def validate_turn_api_token(cls, v: str) -> str:
         return v
 
 
+class Journey(BaseModel):
+    name: Annotated[str, Field(description=("The name of the Turn.io journey, as defined in the Turn.io platform."))]
+    uuid: Annotated[
+        str,
+        Field(description=("The unique identifier (UUID) of the Turn.io journey, as defined in the Turn.io platform.")),
+    ]
+    model_config = ConfigDict(extra="allow")
```
dctaf: Unless you are intentionally consuming extra fields from Journey, consider extra='ignore' instead. `ignore` is also the default, which allows you to omit model_config entirely.

dctaf (approved):

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api_types.py
poornimaramesh: fixed, thanks!

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
poornimaramesh: added sorting by UUID, thanks

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
poornimaramesh: renamed, thanks!

PR #282, opened 2026-06-09
Copyedits to Experiments API documentation.

dctaf:
Improves consistency of terminology and tone used in the public API
documentation. Also reworks the introduction section to the Experiments API to
better orient users. No behavioral changes to the API.

### DIFF
--- a/src/xngin/apiserver/routers/common_api_types.py
+++ b/src/xngin/apiserver/routers/common_api_types.py
@@ -77,7 +77,7 @@ class DesignSpecMetricBase(ApiBaseModel):
     ] = None
     metric_target: Annotated[
         float | None,
-        Field(description="Absolute target value = metric_baseline*(1 + metric_pct_change)"),
+        Field(description="Absolute target value = metric_baseline * (1 + metric_pct_change)."),
     ] = None
 
     # Cluster randomization design parameters (all must be set together, or none).
@@ -87,7 +87,7 @@ class DesignSpecMetricBase(ApiBaseModel):
     ] = None
     avg_cluster_size: Annotated[
         float | None,
-        Field(description="Average number of individuals per cluster."),
+        Field(description="Average number of participants per cluster."),
     ] = None
     cv: Annotated[
         float | None,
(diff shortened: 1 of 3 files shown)


src/xngin/apiserver/routers/common_api_types.py
```diff
@@ -175,58 +178,48 @@ def check_has_only_one_of_pct_change_or_target(self) -> Self:
...
 
 
 class OnlineAssignmentWithFiltersRequest(ApiBaseModel):
-    """Request model for creating a new frequentist online assignment with server-side filtering.
-
-    Usage notes:
-    1. Currently only used for FREQ_ONLINE experiments.
-    2. If the experiment defines no filters, use the corresponding GET endpoint instead.
-    3. If an assignment already exists for a given participant, the property list is ignored and assignment returned.
-    4. Property names must reference a valid field_name from the experiment's fields.
```
poornimaramesh, line 193: the copy-edit seems to have removed these additional instructions about the filter inputs that are not captured in the endpoint description. Maybe we can add these back in?

src/xngin/apiserver/routers/common_api_types.py
```diff
@@ -650,37 +633,37 @@ class MetricPowerAnalysis(ApiBaseModel):
...
+                "The minimum detectable effect (MDE) achievable for design_spec.desired_n at the chosen "
+                "confidence and power. Present only when design_spec.desired_n is set (frequentist design specs)."
             )
         ),
     ] = None
 
     msg: Annotated[
         MetricPowerAnalysisMessage | None,
-        Field(description="Human friendly message about the above results."),
+        Field(description="A human-friendly message describing these results."),
```
poornimaramesh: "these results" feel quite imprecise. Maybe we can update it to say "power analysis results"?

src/xngin/apiserver/routers/common_api_types.py
```diff
@@ -1055,8 +1034,10 @@ def check_contexts(self) -> Self:
 
 
 class PreassignedFrequentistExperimentSpec(BaseFrequentistDesignSpec):
-    """Use this type to randomly select and assign from existing participants at design time with
```
poornimaramesh, line 1058: we should mention that this still works with an _existing_ cohort of users -- the copy-edit seems to have removed that bit (or at least it's not clear anymore)

poornimaramesh (approved):
Looks good to me overall, thanks so much! The docstrings do seem to read a lot more nicely now :)

src/xngin/apiserver/routers/common_api_types.py
dctaf: Oops, I should have left a comment about this. I intentionally removed these for a few reasons:

* The existing comments muddled API endpoint documentation with payload body documentation,
* The existing docs are imprecise,
* experiment_id-based filter support was removed a few months ago,
* @markbotterill 's work will change this behavior, so I believe we will end up with better docs if they are written by our new domain expert.

src/xngin/apiserver/routers/common_api_types.py
dctaf: Done

src/xngin/apiserver/routers/common_api_types.py
dctaf: Good point! I've reworded this. Clarity on whether users exist before experiment design or are assigned after experiment design is something we will probably want to communicate earlier in the documentation as well.

PR #283, opened 2026-06-09
Handle check_power() errors in MDE mode

qixotic:
No longer swallows the errors, but rather propagates them up to the outer try/except, so that it can explicitly handle some internal statsmodels errors and return friendlier messages.

Also refactors StatsPowerError to allow those custom messages, and the top level exception handler to display the specific StatsError class name instead of just a generic (stats).

## Ticket

Fixes #EVE-34

## Related PRs

https://github.com/agency-fund/evidential-fe/pull/262

## How has this been tested?

unittest updates

### DIFF
--- a/src/xngin/stats/power.py
+++ b/src/xngin/stats/power.py
@@ -145,25 +145,29 @@ def check_power(
 
             # Optional desired_n MDE calculation added to the min sample size analysis:
             if desired_n is not None:
-                try:
-                    mde_analysis = analyze_metric_power(
-                        metric=metric,
-                        n_arms=n_arms,
-                        arm_weights=arm_weights,
-                        desired_n=desired_n,
-                        power=power,
-                        alpha=alpha,
-                    )
-                    # pct_change_possible may be None if there was any error created with
-                    # power_analysis_error().
-                    if mde_analysis.pct_change_possible is not None:
-                        analysis.pct_change_with_desired_n = mde_analysis.pct_change_possible
-                except ValueError:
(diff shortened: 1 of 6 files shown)


src/xngin/stats/power.py
```diff
@@ -145,25 +145,29 @@ def check_power(
...
+                )
+                # pct_change_possible may be None if there was any error created with
+                # power_analysis_error().
+                if mde_analysis.pct_change_possible is not None:
+                    analysis.pct_change_with_desired_n = mde_analysis.pct_change_possible
 
             analyses.append(analysis)
+
+        except ZeroDivisionError as zde:
+            raise StatsPowerError("Zero division - You likely have too few samples.") from zde
```
dctaf, line 164: Here's a rewrite that applies the proposed product copy guidelines:

```
"The sample size is too small to compute power. Increase the sample size and try again."
```

src/xngin/stats/power.py
```diff
@@ -145,25 +145,29 @@ def check_power(
...
         except ValueError as verr:
-            raise StatsPowerError(verr, metric.field_name) from verr
+            if "f(a) and f(b) must have different signs" in str(verr):
+                raise StatsPowerError(
+                    "Unable to solve. Check your sample size, effect size, or significance parameters."
```
dctaf, line 168: "Could not compute power with the given inputs. Update the sample size, effect size, or significance, then try again."

dctaf (approved):
Please consider incorporating the copy feedback below.

PR #286, opened 2026-06-15
Avoid DNS resolvers during unit tests.

dctaf:
Some of the delays we saw in the meeting today are consistent with DNS issues.
This PR attempts to eliminate live DNS dependencies from unit tests, and also cleans
up some correctness issues in the implementation of `safe_resolve.py`.


src/xngin/apiserver/dns/test_safe_resolve.py
```diff
@@ -0,0 +1,83 @@
...
+    ("1.2.3.4", True),
+    ("10.0.0.1", False),
+    ("100.64.0.1", False),
+    ("127.0.0.1", False),
+    ("169.254.169.254", False),
+    ("172.16.0.1", False),
+    ("192.168.1.1", False),
+    ("192.30.252.1", True),
+    ("8.8.8.8", True),
+    ("not-an-ip", False),
```
qixotic, line 37: how about also adding a public multicast and v6 ip each for your direct test of is_safe_ip?

qixotic (approved):

src/xngin/apiserver/dns/test_safe_resolve.py
dctaf: Done.

PR #287, opened 2026-06-18
CI job checking for breakages to frontend build

dctaf:
Posts a non-blocking comment on PRs indiciating how the change will affect the
frontend build.  This is mostly vibed.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -154,6 +154,7 @@ jobs:
       pydependencies: ${{ steps.filter.outputs.pydependencies }}
       migrations: ${{ steps.filter.outputs.migrations }}
       queries: ${{ steps.filter.outputs.queries }}
+      openapi: ${{ steps.filter.outputs.openapi }}
     steps:
       - if: github.event_name != 'pull_request'
         uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -176,7 +177,10 @@ jobs:
             queries:
               - 'src/xngin/apiserver/dwh/**'
               - 'src/xngin/apiserver/conftest.py'
-
+            openapi:
+              - 'src/**/*.py'
```
poornimaramesh, line 181: should ew restrict this to changes on **/\*_types.py and **/\*_api.py?

poornimaramesh (approved):
Looks good and approving. 

More generally, I am wondering if the pull from evidential-fe and running orval is overkill, if it's a non-blocking CI check? Maybe it's enough to flag that the OpenAPI spec has changed and that we should run `task genapiclient` on the frontend -- assuming that's the behaviour we want to encourage anyway.

Or maybe it's time to switch to a monorepo ;)

.github/workflows/test.yaml
dctaf: Perhaps. That would save CI time, but we don't enforce that the content in the OpenAPI spec only comes from those files. For example, our customizations in openapi.py also influences the output. Let's see how this plays out and we can refine as needed.

dctaf:
> Looks good and approving.
> 
> More generally, I am wondering if the pull from evidential-fe and running orval is overkill, if it's a non-blocking CI check? Maybe it's enough to flag that the OpenAPI spec has changed and that we should run `task genapiclient` on the frontend -- assuming that's the behaviour we want to encourage anyway.

This CI step is fragile and will break more often than developers make mistakes. I would like it to stay informational so that we don't end up encouraging the bad habit of overriding a flaky step or due to deployment realities. 

Moving to a monorepo is probably the best future.

Thanks!

PR #292, opened 2026-06-21
Add cluster_counts info to AssignSummary's ArmSize [EVE-130]

qixotic:
## Description

This change is scoped just to cluster randomized experiments:
* On experiment creation, convert_assignment_results_to_assign_summary() fills cluster_count from the new in-memory AssignmentResult.arm_cluster_pop.
* On experiment get/list, when the experiment has a cluster_key, get_assign_summary() runs a COUNT(DISTINCT cluster_key) grouped by the arm ids.

This is in support of https://github.com/agency-fund/evidential-be/issues/217, which asks to display true cluster counts (not approximate ones as output by a power check) as part of Arms & Allocations info.

## Future Tasks (optional)

If performance is an issue for future clustered experiments, so long as these experiments are restricted to preassigned-only, we can create an ArmClusterCounts table that stores the AssignmentResult.arm_cluster_pop counts on experiment creation. 

## How has this been tested?

unittests, including comparisons of creation and fetch-time


src/xngin/apiserver/routers/experiments/experiments_common.py
```diff
@@ -1074,16 +1084,35 @@ async def get_assign_summary(
...
+            )
+            .outerjoin(
+                tables.ArmAssignment,
+                (tables.Arm.id == tables.ArmAssignment.arm_id) & (tables.ArmAssignment.experiment_id == experiment_id),
+            )
+            .where(tables.Arm.experiment_id == experiment_id)
+            .group_by(tables.Arm.id)
+        )
+        cluster_count_by_arm_id = {arm_id: cluster_count for arm_id, cluster_count in cluster_result}
+
```
dctaf: I am concerned about this query because it will become expensive.

Can we replace this potentially expensive DISTINCT + join usage with a precomputed stat on ArmStats, such as ArmStats.cluster_count, populated from AssignmentResult.arm_cluster_pop in bulk_insert_arm_assignments? If so, we can include it in the earlier query so that we only do one cheap query instead of one cheap query + one potentially expensive query.

dctaf (approved):
Please consider fixing the performance issue (below) prior to merging.

src/xngin/apiserver/routers/experiments/experiments_common.py
qixotic: done

PR #293, opened 2026-06-22
Raise a clear error when a high cluster design effect collapses effective N

mdulieu:
## Problem

For cluster-randomized designs with a very high ICC and few, large clusters, the power check returned misleading errors even when the user entered a large `desired_n`:

- `Power calc error for metric current_income: Chosen sample size must be positive.`
- `Your sample size is too small for the power calculation. Increase it and try again.`

Both tell the user to increase their sample size, which doesn't help — the real driver is the clustering structure.

Repro: standard `a_dwh`, `pk=id`, `clusterkey=ethnicity`, `metric=current_income`, then estimate. 6 natural clusters, avg cluster size ≈ 166,667, ICC ≈ 0.589 → DEFF ≈ 98,227.

## Root cause

`solve_for_mde_cluster_impl` computes `effective_n = int(desired_n / DEFF)` and passes it straight to the individual-power solver. When DEFF is enormous, `effective_n` rounds down to 0/1/2/3, and statsmodels then fails three different ways depending on the exact value (`ValueError: Chosen sample size must be positive`, `ZeroDivisionError`, or `f(a) and f(b) must have different signs`). All three surfaced as "increase your sample size."

## Fix

Add `_check_effective_n_is_usable`, called right after `effective_n` is computed. If the smallest arm would get fewer than 2 effective units, it raises a `ValueError` naming the real cause (DEFF, ICC, average cluster size) and noting that a larger sample won't meaningfully help. It's a plain `ValueError` so it reuses the existing `check_power` wrapper and keeps the metric name in the message. The guard lives in `solve_for_mde_cluster_impl`, the choke point shared by both the MDE and sample-size code paths.

Users now see, e.g.:

> Power calc error for metric current_income: The design effect from clustering (DEFF=98227) shrinks the effective sample size to 0 (from a desired 50000), which is too small to run a power calculation. This is driven by a high intracluster correlation (ICC=0.5894) combined with a large average cluster size (166667) spread across few clusters, not by the desired sample size. Increasing the desired sample size will not meaningfully help; reduce the ICC or use more (and smaller) clusters.

## Tests

- Parametrized regression test over the four collapsing `desired_n` values (effective_n 0/1/2/3), asserting the clear message and the precise per-case effective/desired N.
- Success-boundary test pinning exact DEFF, effective_n, target_possible, and pct_change_possible once effective_n is usable.
- End-to-end test through `check_power` confirming the user-facing wrapper keeps the metric name and drops the old misleading wording.


qixotic (approved):
LGTM after I tweaked the error message further and cleaned up some extra wordy comments. Thanks!

PR #294, opened 2026-06-23
Add inbound webhook for Turn.io app to trigger a journeys refresh on Evidential

poornimaramesh:
## Ticket

[EVE 126](https://idinsight.atlassian.net/browse/EVE-126?atlOrigin=eyJpIjoiNmJmMWRkOTUyOTQ3NDhjODhkNWQ4MWZiY2IwNzI5ODUiLCJwIjoiaiJ9)

## Related PRs

Frontend: [PR 278](https://github.com/agency-fund/evidential-fe/pull/278)

## Description
- Add a webhook endpoint for the Turn.io App to send requests to whenever Journeys on the Turn.io platform changes
   - reuse the Task and Event tables + handlers to fire off a refresh journeys fetch whenever the webhook endpoint is queried
- Add an endpoint to regenerate a webhook token

## Future Tasks (optional)
- Update Turn App code

## How has this been tested?
- Run tests
- Do `curl` queries for the new endpoints

## Checklist


src/xngin/apiserver/routers/admin/admin_api_types.py
```diff
@@ -274,58 +275,40 @@ class AddWebhookToOrganizationRequest(AdminApiBaseModel):
...
-    @classmethod
-    def validate_url(cls, v: str) -> str:
-        return validate_webhook_url(v)
+    @model_validator(mode="after")
+    def validate_url(self) -> AddWebhookToOrganizationRequest:
+        if self.url is not None:
+            self.url = validate_webhook_url(self.url)
+        if self.url is None and self.type == "experiment.created":
+            raise ValueError("URL must be provided for experiment.created webhooks.")
+        return self
```
qixotic: Shouldn't we add another check that raises when (self.type == "turn.journeys_changed" and direction != "inbound")? And raise if self.url != None for that type?

Actually, given the current restrictions on an experiment.created (outbound w/ url) vs turn.journeys_changed (inbound, url:None = None), I think we could split this into two separate models, then create a discriminated union type like so:

```
AddWebhookToOrganizationRequest = Annotated[
    AddExperimentCreatedWebhookRequest | AddTurnJourneysChangedWebhookRequest,
    Field(discriminator="type"),
]
```
That way a lot of the validation can be moved into the request structure itself.

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
```diff
@@ -231,7 +258,65 @@ async def get_organization_turn_connection(
...
+
+@router.put(
+    "/integrations/turn-connection/{organization_id}/regenerate-webhook-token", responses=TURN_JOURNEYS_RESPONSES
+)
+async def regenerate_turn_webhook_token(
+    organization_id: str,
+    session: Annotated[AsyncSession, Depends(xngin_db_session)],
+    user: Annotated[tables.User, Depends(require_user_from_token)],
+    allow_missing: Annotated[
+        bool, Query(description="If true, return a 200 with null body if the resource does not exist.")
```
qixotic: Wouldn't it be more accurate to say response with the various fields set to the empty string and url None?

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
qixotic: This description is misleading as well; you return a GetTurnConnectionResponse all the time

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
```diff
@@ -231,7 +258,65 @@ async def get_organization_turn_connection(
...
+    turn_webhook = (
+        await session.execute(
+            select(tables.Webhook).where(
+                tables.Webhook.organization_id == org.id, tables.Webhook.type == "turn.journeys_changed"
+            )
+        )
+    ).scalar_one_or_none()
+
+    if turn_webhook is None and not allow_missing:
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="Turn journey changed webhook not found")
```
qixotic: 298-307 could be extracted as a helper function; looks repeated in get_organization_turn_connection and set, although the latter would allow_missing.

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
```diff
@@ -160,7 +161,7 @@ async def set_organization_turn_connection(
     user: Annotated[tables.User, Depends(require_user_from_token)],
     body: Annotated[SetConnectionToTurnRequest, Body(...)],
     httpx_client: Annotated[httpx2.AsyncClient, Depends(retrying_httpx_dependency)],
-):
+) -> AddWebhookToOrganizationResponse:
```
qixotic, line 193: Think it would be cleaner to define a new response type for setting a turn connection in src/xngin/apiserver/routers/admin_integrations/admin_integrations_api_types.py. It could still inherit from AddWebhookToOrganizationResponse, but gives us the flexibility to change things if needed in the future without impacting the core webhook apis.

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
qixotic: Comment needs updating

src/xngin/apiserver/routers/admin_integrations/test_admin_integrations_api.py
qixotic, line 1: Can you update your lifecycle test to also validate the response of set_organization_turn_connection, and verify that the cascading delete works?

src/xngin/apiserver/routers/integrations/integrations_api.py
```diff
@@ -70,3 +72,46 @@ async def get_turn_app_config(
...
+    """
+    webhook = await session.get(tables.Webhook, webhook_id)
+
+    if not webhook:
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="Webhook not found.")
+
+    if auth_token != webhook.auth_token:
+        raise HTTPException(status_code=status.HTTP_401_UNAUTHORIZED, detail="Invalid webhook auth token.")
+
+    # Process the webhook payload here
```
qixotic: Before processing, we should also raise a 404 if that the fetched webhook's type == "turn.journeys_changed" && direction == "inbound", to prevent someone with a different webhook_id && auth_token from arbitrarily queuing new TURN_JOURNEYS_CHANGED_TASK_TYPEs for the org.

qixotic: or fetch with
```
webhook = (
    await session.execute(
        select(tables.Webhook).where(
            tables.Webhook.id == webhook_id,
            tables.Webhook.type == "turn.journeys_changed",
            tables.Webhook.direction == "inbound",
        )
    )
```

src/xngin/apiserver/routers/integrations/integrations_api.py
```diff
@@ -70,3 +72,46 @@ async def get_turn_app_config(
...
+    session.add(
+        tables.Task(
+            task_type=TURN_JOURNEYS_CHANGED_TASK_TYPE,
+            payload={"organization_id": webhook.organization_id},
+        )
+    )
```
qixotic, line 129: Wouldn't you want to also insert a analogous to what we do in experiments_common.py::commit_experiment_impl()? Then we could display a message about the refresh of journeys in the org's Recent Events section.

Ah nevermind I see it's done after the actual refresh in the tq handler. 👍

src/xngin/events/turn_journeys_changed.py
```diff
@@ -0,0 +1,19 @@
...
+class TurnJourneysChangedEvent(BaseEventModel):
+    """Describes the result of an inbound webhook request."""
+
+    TYPE: ClassVar[Literal["turn.journeys_changed"]] = "turn.journeys_changed"
+
+    type: Literal["turn.journeys_changed"] = "turn.journeys_changed"
+    organization_id: Annotated[str, Field(description="The organization ID.")]
+    success: bool
+    response: str
+
```
qixotic, line 18: Can you also add a 

    def status_icon(self) -> Literal["success", "failure"]:
        return "success" if self.success else "failure"

like WebhookSentEvent does?

src/xngin/tq/handlers.py
```diff
@@ -132,3 +157,69 @@ def webhook_outbound_handler(task: Task):
...
+                        session,
+                        request,
+                        success=False,
+                        response_summary=f"No Turn connection found for organization {request.organization_id}",
+                    )
+                    raise ValueError(f"No Turn connection found for organization {request.organization_id}")
+
+                transport = httpx2.AsyncHTTPTransport(retries=3)
+                async with httpx2.AsyncClient(transport=transport) as client:
+                    journeys = await refresh_journeys_dict(turn_connection.get_turn_api_token(), client)
```
qixotic: This calls secretservice.get_symmetric().decrypt(), so we'll need to add secretservice.setup() to src/xngin/tq/cli.py.

I discovered this by manually triggering a turn journeys changed task:
```
curl -X POST "http://localhost:8000/v1/integrations/turn/webhook/wh_dP4vk2MCRSi9x52z" \ 
     -H "Webhook-Token: TOKEN"
```
and `task start-tq` yielding error:
`xngin.xsecrets.exceptions.InvalidSecretStoreConfigurationError: setup() must be called before encryption operations`

The tests didn't catch this since they indirectly load the main app (which does the init) via the conftest.py loaded in test_handlers.py with:
`pytest_plugins = ("xngin.apiserver.conftest",)`

src/xngin/apiserver/routers/integrations/integrations_api.py
```diff
@@ -70,3 +72,46 @@ async def get_turn_app_config(
...
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="Webhook not found.")
+
+    if auth_token != webhook.auth_token:
+        raise HTTPException(status_code=status.HTTP_401_UNAUTHORIZED, detail="Invalid webhook auth token.")
+
+    # Process the webhook payload here
+    session.add(
+        tables.Task(
+            task_type=TURN_JOURNEYS_CHANGED_TASK_TYPE,
+            payload={"organization_id": webhook.organization_id},
```
qixotic: Let's validate first the payload in case the model ever changes:
```
payload=TurnJourneysChangedTask(organization_id=webhook.organization_id).model_dump(),
```

src/xngin/tq/test_handlers.py
```diff
@@ -120,3 +126,99 @@ def handler(request: httpx2.Request) -> httpx2.Response:
...
+            task_type=TURN_JOURNEYS_CHANGED_TASK_TYPE,
+            payload={"organization_id": org_id},
+        )
+        success_task = await wait_for_task_status(task.id, "success")
+
+    assert success_task.message is None
+
+    # Journeys dict should reflect the latest stacks response.
+    turn_connection = (
+        await xngin_session.execute(
```
qixotic: Rather than query the db directly, how about using

    journeys = iaclient.get_organization_turn_journeys(organization_id=org_id).data.journeys

since we've been trying to minimize tests that hit the db directly if we can use an api.

qixotic (changes requested):
Cool feature! Have a few suggested changes and fixes, but they should all be straightforward.

qixotic:
BTW, does this sequence diagram look correct for intended interaction:

```mermaid
sequenceDiagram
    participant Admin
    participant Evidential
    participant Turn

    rect rgba(253, 253, 240)
        Note over Admin,Turn: Initial Setup
        Admin->>Evidential: PUT turn-connection (Turn API token)
        Evidential-->>Admin: webhook id + auth_token (show once)
        Admin->>Turn: Configure webhook URL + Webhook-Token
    end

    Turn->>Evidential: POST /integrations/turn/webhook/{id}<br/>Header: Webhook-Token
    Evidential->>Evidential: Validate token, enqueue task
    Evidential->>Turn: GET journeys (using turn_api_token)
    Evidential->>Evidential: Update journeys_dict
```

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
qixotic: After looking at the FE UI, I think this response and  GetTurnConnectionResponse from the getter below need to be aligned more. (If they could be the same response schema, even better.)

GetTurnConnectionResponse needs at least the webhook id as well to display and copy (currently blank), and why not send the full auth_token and drop auth_token_preview? The UI has a button suggesting I can copy the full string, which would be consistent with how webhooks behave already in the org settings screen. That would allow the UI to still mask everything except the last 4 while allowing you to wire up the copy button.

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api_types.py
```diff
@@ -53,6 +53,15 @@ class GetTurnConnectionResponse(AdminApiBaseModel):
             )
         ),
     ]
+    auth_token_preview: Annotated[
```
qixotic: This should probably be replaced with the real auth_token if you want users to copy still in the FE. Also, can you name it to webhook_auth_token to differentiate from the Turn.io API token_preview? The descriptions for this and token_preview are also a little too similar at first glance and can be confusing.

qixotic (changes requested):

src/xngin/apiserver/routers/integrations/integrations_api.py
```diff
@@ -70,3 +72,46 @@ async def get_turn_app_config(
         experiment_name=experiment.name,
         arm_journey_map=turn_config.arm_journey_map,
     )
+
+
+@router.post(
+    "/integrations/turn/webhook/{webhook_id}",
```
dctaf: Consider adding a collection to specify the type of webhook this is; e.g. `/integrations/turn/webhook/{webhook_id}/config-updated`.

src/xngin/apiserver/routers/integrations/integrations_api.py
```diff
@@ -70,3 +72,46 @@ async def get_turn_app_config(
...
+            "model": dict,
+            "description": "The provided Webhook-Token does not match the token configured for this webhook.",
+        },
+        "404": {
+            "model": dict,
+            "description": "The specified webhook was not found.",
+        },
+    },
+)
+async def turn_webhook(
```
dctaf: These method names are used as operationId in the generated OpenAPI specs. Many code generators use the operationId as the method name in the generated SDKs. I recommend using a more descriptive name for this method.

src/xngin/apiserver/routers/integrations/integrations_api.py
dctaf: Nit: remove low-value comment

src/xngin/tq/handlers.py
dctaf: The reason for this surprise is that tq was designed anticipating that we would eventually move to a [transactional outbox](https://microservices.io/patterns/data/transactional-outbox.html) pattern with a more mature task queue system, and that tq would stay uninformed about the details of the evidential-be business logic. It intentionally avoids any dependencies on sqla.tables, the core business logic of evidential-be, and is meant to be run with fewer privileges. If we add secretservice to tq, then we will increase its operational complexity such that it will more closely resemble the evidential-snapshots service than a simple "fire and forget" service.

Here's an alternative to consider: Instead of editing the TurnConnection row directly, the job can call back into the evidential-be API and ask it to update the turn connections. `set_organization_turn_connection` already implements the refresh behavior, so the only trick is adding the webhook auth token to the Task payload and a small refactoring to allow `set_organization_turn_connection` to accept that token.

If you would prefer to proceed with the PR as designed, that's acceptable too, but it will also invite some ops work:
 
1. Setting the XNGIN_SECRETS variables on the evidential-tq job. The service will fail to start unless these are set. Because these are sensitive values, they are only set on services that **need** them. 
2. Create a tracking ticket to merge evidential-tq and evidential-snapshots job into a single service. This will keep the stack as simple as possible -- once tq brings in all the backend behavior, it has the same operational concerns that evidential-snapshots does, so we can run them in the same service.

dctaf (changes requested):

src/xngin/apiserver/routers/admin/admin_api_types.py
poornimaramesh: Good point -- I've split up the models for outbound webhooks and Turn.io webhooks now, thanks!

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
poornimaramesh: done, thanks!

src/xngin/events/turn_journeys_changed.py
poornimaramesh: done, thanks!

src/xngin/apiserver/routers/integrations/integrations_api.py
poornimaramesh: done, thanks!

src/xngin/tq/test_handlers.py
poornimaramesh: fixed, thanks

src/xngin/apiserver/routers/integrations/integrations_api.py
poornimaramesh: updated, thanks!

src/xngin/apiserver/routers/integrations/integrations_api.py
poornimaramesh: I've updated the name to be more descriptive now, thanks for flagging

src/xngin/apiserver/routers/integrations/integrations_api.py
poornimaramesh: removed, thanks!

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
poornimaramesh: updated, thanks for pointing this out

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api.py
poornimaramesh: Updated!

poornimaramesh: Updated, thanks for flagging!

src/xngin/apiserver/routers/admin_integrations/test_admin_integrations_api.py
poornimaramesh: yes, I've done this now.

src/xngin/apiserver/routers/integrations/integrations_api.py
poornimaramesh: thanks, I've updated the fetch to also filter on webhook type and direction

src/xngin/apiserver/routers/admin_integrations/admin_integrations_api_types.py
poornimaramesh: Okay, thanks! I've updated things so `GetTurnConnectionResponse` now extends `AddWebhookToOrganizationResponse`, so both share id, type, direction, name, url, and auth_token from the webhook — GetTurnConnectionResponse just adds `turn_api_token_preview` on top.
The GET endpoint (id=webhook.id), also returns the webhook ID and the full webhook auth token now, and the FE takes care of masking, same as for the other webhook tokens.

src/xngin/tq/handlers.py
poornimaramesh: Thanks both! I suspected that the current task scheduling system was meant to be a placeholder until we got to a more mature system -- thanks for testing and explaining more clearly!

Here's what I've implemented:

1. New endpoint in `integrations_api.py` that also accepts the webhook ID and token, and triggers the refresh / fetch from the Turn.io. The task handler now directly calls this function. I chose to do this instead of updating `set_organization_turn_connection` because 
(a) this method sits within admin APIs and requires session tokens and requires a Turn API token in the body, which in turn would require the task handler to keep track of internal business logic -- this goes against Doug's advice to keep the fire and forget structure, and 
(b) I think it makes sense to keep the logic for setting a Turn connection separate from the refresh logic. Refreshing the journeys (to me) is a logical side-effect of updating the Turn connection, but we might want to move it out of that endpoint in the future, and it would be nice not to have to touch the webhook task handling then

2. Some awkward design things: we now effectively have 2 endpoints in the external facing API: one to trigger the refresh task, and the other to actually perform the task. We can't unify them right now without affecting performance (Turn.io times out webhooks after 5s, and it doesn't look like we can change that limit, so if there a lot of Journeys on the platform, the refresh will take a while), but now we have this weird loopback between the endpoints and the task handler.

I'm ok to keep it as is, since it accomplishes what we want, and make a tracking ticket so we can revisit this decision in the future, but let me know if you have other ideas.

src/xngin/apiserver/routers/integrations/integrations_api.py
```diff
@@ -70,3 +84,103 @@ async def get_turn_app_config(
...
+    It is not intended to be called directly by clients, and will return a 400 error if called without
+    a valid Turn.io webhook auth token.
+
+    This endpoint only enqueues a task; it does not perform the refresh itself. A `tq` worker picks up
+    the task and triggers the actual refresh by calling `refetch_journeys_from_turn` below, over HTTP.
+    This is intentional: Turn.io's webhook calls timeout in ~5s, and large refreshes may take longer.
+    See `make_turn_journeys_changed_handler` in `xngin/tq/handlers.py` for the worker side logic.
+    """
+    if not auth_token:  # for mypy
+        raise HTTPException(status_code=status.HTTP_401_UNAUTHORIZED, detail="Missing webhook auth token.")
```
qixotic: This check is redundant with the call to `check_webhook_auth_token` below. I'd opt for code simplicity over this little early exit, or just move the call to `get_turn_webhook_or_raise` into `check_webhook_auth_token` and return appropriately with a function name update.

src/xngin/apiserver/routers/integrations/integrations_api.py
```diff
@@ -70,3 +84,103 @@ async def get_turn_app_config(
...
+    """
+    Refreshes the cached Turn.io journeys for the organization owning this webhook.
+
+    This endpoint is called in two ways:
+    1. By the `tq` worker's `turn_journeys_changed` handler, in response to a Turn.io webhook
+       notification (see `receive_turn_journey_update_notification` above). The worker calls this
+       endpoint over HTTP, authenticating with the same Webhook-Token any other caller would use,
+       rather than invoking the refresh logic in-process — `tq` intentionally has no dependency on
+       Evidential's business logic or database models, so it treats this as an external API call.
+    2. Directly, as a way to manually trigger a refresh without waiting for a Turn.io webhook or
```
qixotic, line 159: This 2nd case can be deleted right (and tweak the doc further up too)? `set_organization_turn_connection` calls `refresh_journeys_dict` directly.

src/xngin/tq/handlers.py
```diff
@@ -132,3 +160,62 @@ def webhook_outbound_handler(task: Task):
...
+            raise ValueError("Task payload is empty")
+
+        request = TurnJourneysChangedTask.model_validate(task.payload)
+        logger.info(f"Processing request for org {request.organization_id} and webhook {request.webhook_id}")
+        with Session(engine) as session:
+            try:
+                async with httpx2.AsyncClient(transport=transport, timeout=10.0) as client:
+                    response = await client.request(
+                        method="POST",
+                        url=f"{XNGIN_PUBLIC_API_BASE_URL}{constants.API_PREFIX_V1}/integrations/turn/webhook/{request.webhook_id}/refresh-journeys",
```
qixotic, line 197: Don't forget to set XNGIN_PUBLIC_API_BASE_URL on the evidential-tq job in railway

src/xngin/tq/handlers.py
qixotic: Makes sense. I'd also note in any new tracking ticket that this logic also keeps a copy of plaintext webhook_auth_token in the task payload (which today is our tables.Task) as another place to keep in mind in case we wish to one day lock that down further somehow.

qixotic (approved):
Just a little more cleanup. Thanks!

src/xngin/apiserver/routers/integrations/integrations_api.py
poornimaramesh: Yes, but only when you rotate the key. If you want to manually trigger a refresh without resetting the key, we still need to hit this endpoint.

src/xngin/apiserver/routers/integrations/integrations_api.py
poornimaramesh: removed, thanks!

src/xngin/tq/handlers.py
poornimaramesh: done, thanks for flagging!

poornimaramesh: also added the ticket for the plaintext [webhook token](https://idinsight.atlassian.net/browse/EVE-150?atlOrigin=eyJpIjoiMzAyZmJjMDI2NTAzNGQ3YzgzNjk1ZTViYWFjMjk5NWIiLCJwIjoiaiJ9)

src/xngin/apiserver/routers/integrations/integrations_api.py
poornimaramesh: I'll leave it as is for now, but made a ticket to track it: https://idinsight.atlassian.net/browse/EVE-151?atlOrigin=eyJpIjoiOTE4OWM2MjYyM2ViNDg1MGFjYzZhZWQ0M2Q4MWVkZjUiLCJwIjoiaiJ9

PR #263, opened 2026-05-22
Add clustered assignment support to preassigned freq experiment creation.

qixotic:
## Description

Wire cluster_key into preassigned experiment assignment.
Also add extra validator on BaseFrequentistDesignSpec to ensure new experiment requests don't accidentally set strata on cluster designs for now.

Followup to #163 as part of issue #217.

## Future Tasks

Support analysis of clustered designs in API endpoint.

## How has this been tested?

unittests

PR #295, opened 2026-06-23
Fix cluster assignment

qixotic:
## Description

Assignment bug: For clustered experiments, we were incorrectly randomly sampling the _individual_ desired_n participants from the dwh, and then looking at what clusters they happened to be associated with. (This was an oversight when wiring up the API in #263.)

Fix: add a PreassignedFrequentistExperimentSpec.desired_n_clusters field and persist it in the table Experiments. Add a new DwhSesion.get_clusters_of_participants() and helpers to sample the number of clusters (instead of the number of participants). Since participant counts aren't guaranteed as entire clusters are assigned to arms, some tests needed updating as well to verify expected cluster counts.

related: https://github.com/agency-fund/evidential-fe/pull/272

## Future Tasks

* update the FE code to use this new interface
* consider updating the power check interface to also respect the new desired_n_clusters

### DIFF
--- a/src/xngin/apiserver/dwh/dwh_session.py
+++ b/src/xngin/apiserver/dwh/dwh_session.py
@@ -8,7 +8,7 @@ from typing import Self
 import google.api_core.exceptions
 import sqlalchemy
 from loguru import logger
-from sqlalchemy import Engine, Inspector, event, text
+from sqlalchemy import ColumnElement, Engine, Inspector, event, text
 from sqlalchemy.engine.interfaces import DBAPIConnection
 from sqlalchemy.exc import NoSuchTableError, OperationalError
 from sqlalchemy.orm import Session
@@ -272,7 +272,7 @@ class DwhSession:
         self,
         table_name: str,
         filters: list[Filter],
-        compose_query: Callable[[sqlalchemy.Table, list[Filter]], sqlalchemy.Select],
+        compose_query: Callable[[sqlalchemy.Table, list[ColumnElement]], sqlalchemy.Select],
         use_sa_autoload: bool | None = None,
     ) -> GetParticipantsResult:
         """Inspects ``table_name``, runs the query built by ``compose_query``, and wraps the resulting rows.
@@ -284,7 +284,7 @@ class DwhSession:
         sa_table = self._inspect_table_blocking(table_name, use_sa_autoload=use_sa_autoload)
         sqla_filters = query_constructors.create_query_filters(sa_table, filters)
         participants = self.session.execute(compose_query(sa_table, sqla_filters)).all()
-        return GetParticipantsResult(sa_table=sa_table, participants=participants)
+        return GetParticipantsResult(sa_table=sa_table, participants=list(participants))
 
     def _get_participants_blocking(
         self,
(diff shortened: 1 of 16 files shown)


src/xngin/apiserver/routers/experiments/experiments_common.py
```diff
@@ -243,9 +243,13 @@ async def create_experiment_impl(
     match request.design_spec:
         case PreassignedFrequentistExperimentSpec():
             preassigned_spec = request.design_spec
+            cluster_key = preassigned_spec.cluster_key
             desired_n = preassigned_spec.desired_n
-            if desired_n is None:
-                raise LateValidationError("Preassigned experiments must have a desired_n.")
+            desired_n_clusters = preassigned_spec.desired_n_clusters
+            if cluster_key is not None and desired_n_clusters is None:
+                raise LateValidationError("Cluster-randomized preassigned experiments must have a desired_n_clusters.")
```
poornimaramesh, line 250: this case is already caught by the pydantic validation right?

src/xngin/apiserver/routers/experiments/experiments_common.py
```diff
@@ -243,9 +243,13 @@ async def create_experiment_impl(
...
+            if cluster_key is None and desired_n is None:
+                raise LateValidationError("Individual-randomized preassigned experiments must have a desired_n.")
```
poornimaramesh, line 253: we could add this check as a validator method on the `PreassignedFrequentistSpec` pydantic model, no?

src/xngin/apiserver/dwh/dwh_session.py
```diff
@@ -287,9 +294,34 @@ def _get_participants_blocking(
...
+        desired_n_clusters: int,
+        cluster_key: str,
+        use_sa_autoload: bool | None = None,
+    ) -> GetParticipantsResult:
+        return self._get_result_blocking(
+            table_name,
+            filters,
+            lambda sa_table, sqla_filters: query_constructors.compose_cluster_query(
+                sa_table,
+                select_columns | {cluster_key},
```
poornimaramesh, line 318: nit: `select_columns` already has `cluster_key` included when we call this function.

poornimaramesh (approved):
Looks good (added some minor comments), tests pass, approving, thanks!

src/xngin/apiserver/routers/experiments/experiments_common.py
qixotic: PreassignedFrequentistExperimentSpec only checks that `desired_n_clusters` is never set if it's not clustered.

I haven't moved the validation up into the spec because it's reused for power calculations, where desired_n (and in the future desired_n_clusters) may not be set when it's used for a min sample size calculation.

That said, I've created EVE-135 to think about some validation cleanup, thanks.

src/xngin/apiserver/routers/experiments/experiments_common.py
qixotic: EVE-135 to handle this better

src/xngin/apiserver/dwh/dwh_session.py
qixotic: Thanks yeah; I'd like to keep it out of paranoia and to be more self-documenting. Thought it looked slightly odd that we add `primary_key` a few lines earlier but not cluster_key, and acknowledges that we do want both when creating an experiment.

PR #296, opened 2026-06-30
DWH-C/MABs Phase 2: Backend for Sample Calls Logic

markbotterill:
## Ticket
[Epic](https://idinsight.atlassian.net/browse/EVE-8)

## Related PRs
Frontend: `mab-phase2-sample-calls-fe` (renders these calls in the integration panel) - see [here](https://github.com/agency-fund/evidential-fe/pull/281)

## Description

Adds a `sample_calls` field to the admin `get_experiment_for_ui` response. There are structured, type-correct example API calls so integrators can see exactly what to send and expect as s response. We provide examples for all the GET assignment calls (0/1 for a Bernoulli reward or boolean DWH target, else an illustrative number)+ the POST responses where possible (this applies for the MAB/ MAB-DWH)

## NBs:

1. CMAB is currently unimplemented but will be added once CMAB DWH progresses. 

2. Additive and backwards-compatible so public `/v1/` endpoints are untouched.

3. Examples are filled with the experiment id and a real (baseline) arm; `participant_id`and the API key stay as placeholders.

## How has this been tested?
  - Unit tests for `make_sample_calls` (a parametrized per-type contract table + structural, outcome-value, and real-arm tests).
  - API-level tests in `test_admin_api.py`: MAB (both calls), frequentist (get-assignment only), CMAB (`null`).

  ## Checklist


src/xngin/apiserver/routers/experiments/experiments_common.py
```diff
@@ -908,6 +910,86 @@ async def create_assignment_for_participant(
...
+    # A participant is assigned exactly one arm, so show a real one to make the example concrete.
+    # arms are ordered baseline-first, so arms[0] (the control) is the natural illustrative value;
+    # fall back to placeholders if arms aren't loaded.
+    example_arm = experiment.arms[0] if experiment.arms else None
+    arm_id_example = example_arm.id if example_arm else "<arm_id>"
+    arm_name_example = example_arm.name if example_arm else "<arm_name>"
+
+    # Bandits expose the outcome field (filled in via the report-outcome call); for frequentist
+    # experiments the assignment is terminal from the API's perspective, so omit it.
+    assignment_example: dict[str, str | None] = {"arm_id": arm_id_example, "arm_name": arm_name_example}
```
dctaf: Please use the same Pydantic types that we use in the actual API responses when building the assignment_example, the example_responses, and the request bodies. This will ensure that the structures will more match the actual implementation.

src/xngin/apiserver/routers/experiments/experiments_common.py
```diff
@@ -908,6 +910,86 @@ async def create_assignment_for_participant(
...
+                "participant_id": "{participant_id}",
+                "assignment": assignment_example,
+            },
+        ),
+    ]
+
+    if is_bandit:
+        # reward_type is a nullable column (frequentist experiments leave it None), but a bandit
+        # always has one. Asserted here anyway so that mypy can narrow the type for the next block.
+        assert experiment.reward_type is not None, "bandit experiment is missing a reward_type"
```
dctaf: Nit: asserts can introduce subtle bugs when run in certain environments (https://docs.astral.sh/ruff/rules/assert/) and I try to use them as little as possible. You could skip the assert and the explainer comment by replacing `if is_bandit` with `if is_bandit and experiment_reward_type is not None`.

src/xngin/apiserver/routers/experiments/experiments_common.py
```diff
@@ -908,6 +910,86 @@ async def create_assignment_for_participant(
...
+    arm_id_example = example_arm.id if example_arm else "<arm_id>"
+    arm_name_example = example_arm.name if example_arm else "<arm_name>"
+
+    # Bandits expose the outcome field (filled in via the report-outcome call); for frequentist
+    # experiments the assignment is terminal from the API's perspective, so omit it.
+    assignment_example: dict[str, str | None] = {"arm_id": arm_id_example, "arm_name": arm_name_example}
+    if is_bandit:
+        assignment_example["outcome"] = None
+
+    calls = [
```
dctaf, line 956: As implemented, this may inadvertently mislead future clients running freq_online experiments intending to use the online filters endpoint (`assign_with_filters`). If those users called this API, they'll get an assignment without any filtering applied. So, for freq_online, consider expanding this list into two sample calls. This should be fixed before we promote the freq_online filter behavior.

Also, at some point, we should figure out how to incorporate the existing structured OpenAPI documentation on these endpoints into these sample calls.

src/xngin/apiserver/routers/experiments/experiments_common.py
```diff
@@ -908,6 +910,86 @@ async def create_assignment_for_participant(
     )
 
 
+def make_sample_calls(experiment: tables.Experiment) -> SampleCalls | None:
+    """Build example API calls for an experiment, for the admin UI.
+
+    Currently generates samples along the following pattern:
+
+    A/B Preassigned: GET assignment (no POST - outcomes read from the org's DWH)
+    A/B Online: GET assignment (no POST - outcomes read from the org's DWH)
```
dctaf: The sample calls generated by this method do not handle outcomes for A/B tests. I recommend removing at least the "outcomes read" portion. Also, the matrix of supported features per experiment type is complex, so consider eliding this paragraph entirely because it will be redundant with documentation elsewhere. We can structure the code to be easier to be self-explanatory as this method evolves (such as using `match` methods).

dctaf (approved):
Looks good to me. Approving, but please address the comments below before merging.

Also, please remove any tests that are redundant between test_admin_integrations_api and test_experiments_common. The highest value testing we have is API-level testing and it is easy for us to run them. Redundant tests burn engineer time and tokens and can lead to maintaining behavior in lower-level methods that aren't actually reachable from the API.

src/xngin/apiserver/routers/experiments/experiments_common.py
markbotterill: Gotcha - I know the samples don't handle outcomes, I was trying to make that clear with the matrix (i.e. A/B has GET but no POST). That said, as it reads unclearly I will nix the whole block. There are other places in the codebase which make the patter of API calls + experiment types clear.

src/xngin/apiserver/routers/experiments/experiments_common.py
markbotterill: All addressed! Will make sure I carry that forward with other PRs

src/xngin/apiserver/routers/experiments/experiments_common.py
markbotterill: TIL! Thank you - updating now

src/xngin/apiserver/routers/experiments/experiments_common.py
markbotterill: V good catch - I didn't realize we had the filtering functionality! This actually invalidates the matrix I wrote out earlier where I was saying "Freqs only do GETs". I've updated the `make_sample_calls` to handle adding another example in the case where we have Freq w/ filtering (can do both on the FE)

markbotterill:
Removed the duplicated tests above as well @dctaf. Thank you for the thorough review and approval. I'm going to hold fire until @poornimaramesh can review the complementary FE PR as it'd be good to get both in at once.

src/xngin/apiserver/routers/experiments/experiments_common.py
```diff
@@ -908,6 +913,98 @@ async def create_assignment_for_participant(
...
+    # A filtered freq_online experiment must assign via assign_with_filters: the plain GET can't
+    # evaluate filters, so it would assign everyone. (Filterless freq_online assigns fine via GET.)
+    if experiment_type == ExperimentsType.FREQ_ONLINE and experiment.experiment_filters:
+        # One example property per filter; value is a per-participant placeholder.
+        filters_request = OnlineAssignmentWithFiltersRequest(
+            properties=[
+                ParticipantProperty(field_name=ef.field_name, value="<value>") for ef in experiment.experiment_filters
+            ]
+        )
+        calls.append(
```
qixotic: Can you insert to the front of the list, so that it has precedence over the default SampleCall case? That way users with filters see the very likely case they intend to use: a POST with a payload of properties. (Or alternatively as Doug suggested, restructure everything with a match so it's clearer what examples correspond to which type.)

Arguably, the GET example for an experiment with filters should only be used with `create_if_none=false` as a param, too, to avoid assigning an arm.

src/xngin/apiserver/routers/experiments/test_experiments_common.py
```diff
@@ -2812,3 +2813,47 @@ async def test_arm_population_counter(xngin_session, testing_datasource):
...
+    experiment_type = ExperimentsType.MAB_ONLINE_DWH if target_data_type is not None else ExperimentsType.MAB_ONLINE
+    calls = make_sample_calls(
+        _in_memory_experiment(experiment_type, reward_type=reward_type, target_data_type=target_data_type)
+    )
+    assert calls is not None
+    assert calls.calls[1].body == {"outcome": expected_outcome}
+
+
+def test_make_sample_calls_freq_online_without_filters_get_assignment_only():
+    calls = make_sample_calls(_in_memory_experiment(ExperimentsType.FREQ_ONLINE))
```
qixotic, line 2886: can you also make a pair of fake arms and assert that the baseline arm's id is used?

src/xngin/apiserver/routers/experiments/test_experiments_common.py
```diff
@@ -2812,3 +2813,47 @@ async def test_arm_population_counter(xngin_session, testing_datasource):
...
+        (LikelihoodTypes.BERNOULLI, DataType.NUMERIC, 1),  # bernoulli still forces binary
+    ],
+)
+def test_make_sample_calls_outcome_example_value(reward_type, target_data_type, expected_outcome):
+    experiment_type = ExperimentsType.MAB_ONLINE_DWH if target_data_type is not None else ExperimentsType.MAB_ONLINE
+    calls = make_sample_calls(
+        _in_memory_experiment(experiment_type, reward_type=reward_type, target_data_type=target_data_type)
+    )
+    assert calls is not None
+    assert calls.calls[1].body == {"outcome": expected_outcome}
```
qixotic, line 2880: can you also assert that the generic `<arm_id>` is used in the response?

qixotic (approved):
This is a great addition, thanks! Your description needs updating since you've got a new endpoint instead (which I like). Left some more suggestions (in particular the get_assignment_filtered sample) but please merge after.

src/xngin/apiserver/routers/common_api_types.py
```diff
@@ -1492,6 +1492,33 @@ class UpdateBanditArmOutcomeRequest(ApiBaseModel):
...
+        Field(description="Request headers; secrets are placeholders, never real values."),
+    ]
+    body: Annotated[dict | None, Field(description="Example request body, if any.")] = None
+    example_response: Annotated[dict | None, Field(description="Illustrative success response, if any.")] = None
+
+
+class SampleCalls(ApiBaseModel):
+    """The set of example API calls for an experiment, for display in the admin UI."""
+
+    calls: Annotated[list[SampleCall], Field(description="Ordered example calls an integrator would make.")]
```
qixotic, line 1519: re: https://github.com/agency-fund/evidential-fe/pull/281#discussion_r3553144636, the FE assumes each call in this list has a unique label, so how about adding a field validator to check that?

markbotterill: Sorted!

src/xngin/apiserver/routers/experiments/experiments_common.py
markbotterill: Thanks for catching - makes sense to me. Handled in this [commit](https://github.com/agency-fund/evidential-be/pull/296/commits/fac5b17868c9bbea5441261e80a9f8c3d377d6a1). Held off doing the full `case` logic for now and just ensured that this is the first in the list.

PR #285, opened 2026-06-15
Simplify the Taskfile.

dctaf:
- **Disable atlas update notifier.**: This triggers unnecessary DNS calls.
- **Remove `task migration-reset`**: `task drop-databases` is sufficient and safer.

PR #297, opened 2026-07-16
added betterleaks for secrets detection

Snehaaa18:
## Ticket
Issue: #358
Fixes: [EVE-84](https://idinsight.atlassian.net/browse/EVE-84)

## Related PRs
Frontend: [#285](https://github.com/agency-fund/evidential-fe/pull/285)

### Goal
Prevent accidental pushing of secret keys to the public repository by adding Betterleaks as a secret detection tool in both pre-commit hooks and GitHub Actions.

### Changes
- **Pre-commit hook**: Added betterleaks to `.pre-commit-config.yaml`.Scans staged files for secrets on every `git commit`, blocking the commit if anything is detected.
- **Custom rules** (`.betterleaks.toml`)
- **CI workflow** (`.github/workflows/test.yaml`)

## Checklist
- [ ] I have updated the automated tests


.github/workflows/test.yaml
```diff
@@ -34,6 +34,28 @@ env:
...
+      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+      - uses: dortort/betterleaks-action@cd0b905aa4befb8f42f8b1ab98150180c9e77515 # v0.1.0
+        id: betterleaks
+        with:
+          config: .betterleaks.toml
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          fail-on-leak: 'false'
```
markbotterill: Does this mean that it won't fail if we upload secret?

.pre-commit-config.yaml
```diff
@@ -69,3 +69,7 @@ repos:
     rev: 9257c6050c0261b8c57e712f632dc4a8010109a9  # frozen: v1.25.2
     hooks:
       - id: zizmor
+  - repo: https://github.com/betterleaks/betterleaks
```
markbotterill, line 72: Can we make sure we freeze the version as elsewhere in the file please?

.betterleaks.toml
```diff
@@ -0,0 +1,106 @@
...
+keywords = ["XNGIN_SECRETS_GCP_CREDENTIALS"]
+
+[[rules]]
+id = "gcp-kms-key-uri"
+description = "GCP KMS key URI with project details"
+regex = '''gcp-kms://projects/[a-z0-9-]+/locations/[a-z0-9-]+/keyRings/[a-z0-9-]+/cryptoKeys/[a-z0-9-]+'''
+keywords = ["gcp-kms://"]
+
+[[rules]]
+id = "database-url-with-password"
```
markbotterill, line 53: This rule seems to be triggering a lot for me locally when I run it on my laptop. Can you verify it's catching the right things?

.betterleaks.toml
```diff
@@ -0,0 +1,106 @@
...
+
+[allowlist]
+paths = [
+    '''\.git/''',
+    '''node_modules/''',
+    '''\.mypy_cache/''',
+    '''__pycache__/''',
+    '''\.pytest_cache/''',
+    '''\.venv/''',
+    '''migrations/''',
```
markbotterill, line 94: Do we want to allow all of migrations? Could be an area where we accidentally store a key

markbotterill (changes requested):
Looks like a good start - just needs a few changes so that we can see it working properly on GHA etc.

.github/workflows/test.yaml
```diff
@@ -34,6 +34,28 @@ env:
 
 jobs:
 
+  betterleaks:
+    runs-on: ubuntu-latest
```
dctaf, line 38: Please follow the convention on all the other workflows and use `runs-on: ubuntu-24.04` or `runs-on: ubuntu-slim`.

.betterleaks.toml
```diff
@@ -0,0 +1,106 @@
...
+[[rules]]
+id = "xngin-encrypted-value"
+description = "Encrypted secret value (should never appear in plain commits)"
+regex = '''\$encrypted\$:[A-Za-z0-9+/=]+'''
+keywords = ["$encrypted$"]
+
+[[rules]]
+id = "xngin-api-key"
+description = "xngin API key (xat-prefixed)"
+regex = '''xat[a-zA-Z0-9_-]{20,}'''
```
dctaf, line 19: Nit: Use the `xat_` prefix instead of including _ in the [character class]. This is more consistent with how we actually generate the tokens because `xat_` is a constant prefix.

dctaf (changes requested):
Thanks for this! 

Some of these rules might be redundant with betterleak's built-in rules. For example, betterleaks has rules for `gcp-api-key`, `gcp-service-account`, and `gcp-application-default-credentials` (implemented here https://github.com/betterleaks/betterleaks/blob/b9de7fd88f009e9f65f2421370f0d5fbd4b1be5e/cmd/generate/config/rules/gcp.go#L91). Should we use those rules instead of implementing our own?

We should also make sure the cloudflare rules are enabled because some team members will have Cloudflare credentials in their environment (for managing DNS).

.betterleaks.toml
Snehaaa18: Currently the frontend and the backend repos have the same rules, but some of the rules only apply to the frontend repo, so please do let me know if they require any changes.

.github/workflows/test.yaml
Snehaaa18: updated it and made sure that it fails on detecting secrets.

.betterleaks.toml
Snehaaa18: Thank you for pointing it out. I have updated the files and made sure both frontend and backend have consistent rules.

.github/workflows/test.yaml
Snehaaa18: made sure that it `runs-on: ubuntu-24.04`

dctaf (approved):

PR #299, opened 2026-07-17
fix 500 error when frequentist experiment hits update bandit outcome endpoint

poornimaramesh:


markbotterill (approved):
LGTM 🚀 Thanks for catching!

PR #300, opened 2026-07-21
fail if pre-comit hook not installed in your local repo [EVE-32]

qixotic:
Add a check-pre-commit task to ensure that the local git repo hooks are installed.


markbotterill (approved):
Tested locally and works with no issues! Thanks for raising!

PR #301, opened 2026-07-22
catch failed assignments due to bad balance check [EVE-136]

qixotic:
Catches the error `ValueError: r_matrix performs f_test for using dimensions that are asymptotically non-normal` that can occur during a balance check as part of assignment for preassigned freq experiments and returns a friendlier error message.

## How has this been tested?

unittest

## Checklist


poornimaramesh (approved):
Looks good to me. Thanks!

PR #302, opened 2026-08-04
Swap the names on webhook-related response types for clarity.

dctaf:
Rename some types for clarity: WebhookSummary describes the webhook data model, and AddWebhookToOrganizationResponse happens to be an instance of it.


poornimaramesh (approved):
lgtm, approving!

PR #303, opened 2026-08-04
Avoid triggering a PytestCollectionWarning.

dctaf:
pytest will attempt to interpret classes that start with `Test` as a class full
of tests. Our TestExperiment is not a test, so it fails and emits a warning. I
can't come up with any better name for this class, so instead of renaming it,
I've marked it to be ignored by pytest's automatic test discovery.

This silences the following warning:
```
PytestCollectionWarning: cannot collect test class 'TestExperiment' because it has a __init__ constructor (from: src/xngin/apiserver/routers/admin/test_admin_api.py)
@dataclass(slots=True)
```

### DIFF
--- a/src/xngin/apiserver/routers/admin/test_admin_api.py
+++ b/src/xngin/apiserver/routers/admin/test_admin_api.py
@@ -118,6 +118,8 @@ class HasName(Protocol):
 
 @dataclass(slots=True)
 class TestExperiment:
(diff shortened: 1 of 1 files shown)


poornimaramesh (approved):
How about renaming it to DummyExperiment or something like that?

But this seems like a good workaround to me in any case, approving.

PR #305, opened 2026-08-04
Resolve localhost without using DNS resolver.

dctaf:
localhost is a reserved name. Resolvers that forward to an upstream server --
as in containers and most hosted environments -- usually answer it with
NXDOMAIN, so lookup_v4() can fail on some systems.

Our fix is to answer it directly, without using the DNS server.  This is a
no-op unless ALLOW_CONNECTING_TO_PRIVATE_IPS is set: the safety check still
rejects 127.0.0.1 as non-global in production. In development environments,
we allow 127.0.0.1.


Snehaaa18 (approved):
Thankyou for making changes it has resolved the issue that I was facing in my local setup.

PR #306, opened 2026-08-04
Auto fail for bandits

poornimaramesh:
## Ticket

[Issue EVE-80](https://idinsight.atlassian.net/browse/EVE-80)

## Related PRs

Frontend: [PR 292](https://github.com/agency-fund/evidential-fe/pull/292)

## Description
Allow Evidential to autofill outcome values for participants after user-specified time for bandit experiments

This:
- Adds auto-fail options to the `create_experiment` request
- Piggy-backs off the snapshotter cron job to autofill outcome values for experiments periodically

## How has this been tested?
- Added new unittests
- Create a new experiment (via `curl` / backend) with autofail configured, add participants using the `get_assignment` endpoint for the experiment, back-date the assignment in the postgres db, then run `task start-snapshotter` to check that assignment outcome value gets autofilled.

## Checklist


src/xngin/apiserver/sqla/tables.py
```diff
@@ -755,3 +761,33 @@ class Snapshot(Base):
...
+    id: Mapped[str] = mapped_column(primary_key=True, default=autofail_update_id_factory, unique=True)
+    created_at: Mapped[datetime] = mapped_column(server_default=sqlalchemy.sql.func.now())
+    updated_at: Mapped[datetime] = mapped_column(
+        server_default=sqlalchemy.sql.func.now(), onupdate=sqlalchemy.sql.func.now()
+    )
+    status: Mapped[SnapshotStatus] = mapped_column(server_default="pending")
+    # An optional informative message about the state of this task (for example, if a snapshot fails, it might contain
+    # an informative error message).
+    message: Mapped[str | None] = mapped_column()
+    # JSON serialized form of an ExperimentAnalysisResponse. May be null if the snapshot is not yet a success.
```
dctaf, line 786: ExperimentAnalysisResponse is a type that was originally intended to be used in the API responses. We made the questionable short-term decision to also store this same type in the snapshots. This led to a bug where an API change broke the snapshots deserialization. To avoid this happening again, please open a ticket to replace these types with a more stable type that is independent of the API response types.

src/xngin/apiserver/routers/common_api_types.py
```diff
@@ -550,6 +550,16 @@ class BanditExperimentAnalysisResponse(ApiBaseModel):
         int,
         Field(description="The number of outcomes observed for this experiment."),
     ]
+    fraction_automatically_failed: Annotated[
```
dctaf, line 553: This is a new field without a default value. This means deserializing historical snapshots persisted as analysis snapshots might throw exceptions when deserialized, causing API failures that break the UI. The easy solution is to add a default sentinel value (such as 0.0, or allow it to be nullable). The long-term fix is to create distinct types for use in storage so that we aren't persisting API types.

src/xngin/apiserver/routers/experiments/experiments_common.py
```diff
@@ -1409,4 +1441,16 @@ async def _draws_outcome_aggregates(xngin_session: AsyncSession, experiment_id:
             )
         )
     ).one()
-    return DrawsOutcomeAggregates(n_outcomes=n_outcomes, outcome_std_dev=std_dev)
+    fraction_autofailed_outcomes = (
+        await xngin_session.execute(
+            select(func.coalesce(func.avg(tables.Draw.autofailed_outcome.cast(Integer)), 0.0)).where(
+                tables.Draw.experiment_id == experiment_id,
+                tables.Draw.outcome.is_not(None),
```
dctaf, line 1461: I'm not too familiar with this functionality but I'm concerned about the handling of historical draws that have null autofailed_outcomes. avg() will ignore nulls, and the cast will still evaluate to null for those rows, so I think maybe we need another coalesce otherwise historical rows might create an inaccurate denominator:

```python
      func.coalesce(
          func.avg(func.coalesce(tables.Draw.autofailed_outcome.cast(Integer), 0)),
          0.0,
      )
```

Or, consider whether we can do both queries in a single database query, without the nulls posing their subtlety risk -- maybe something like...

```python
  n_outcomes, std_dev, n_autofailed_outcomes = (
      await xngin_session.execute(
          select(
              func.count(tables.Draw.outcome),
              func.coalesce(func.stddev_pop(tables.Draw.outcome), 0.0),
              func.count().filter(tables.Draw.autofailed_outcome.is_(True)),
          ).where(
              tables.Draw.experiment_id == experiment_id,
              tables.Draw.outcome.is_not(None),
          )
      )
  ).one()
```

src/xngin/apiserver/routers/common_api_types.py
```diff
@@ -1034,6 +1072,19 @@ def check_contexts(self) -> Self:
             raise ValueError("Contexts are only applicable for contextual MAB experiments.")
         return self
 
+    @model_validator(mode="after")
+    def check_autofail_params(self) -> Self:
+        """
+        Validate that the autofail parameters are valid.
+        """
+        if (
```
dctaf, line 1080: Do we need additional handling for this case:

```
  MABDwhExperimentSpec(
      reward_type=LikelihoodTypes.NORMAL,
      target_field_name="is_onboarded",  # DWH BOOLEAN column
      enable_autofail=True,
      autofail_outcome_value=0.5,   # <--- not a boolean
      ...
  )
```

I'm concerned this could fail the storage-class aware check in _check_outcome_against_mab_dwh_target later.

(Credit goes to 5.6 Sol for identifying this 🪄)

dctaf (changes requested):

src/xngin/apiserver/sqla/tables.py
poornimaramesh: Thanks for flagging! I've created a ticket [here](https://idinsight.atlassian.net/browse/EVE-165?atlOrigin=eyJpIjoiNDY5ZGQ5M2JlMzUxNDNmMGI4YjhhYjI3NjYyNmUyMDgiLCJwIjoiaiJ9) and will tackle ina separate PR.

src/xngin/apiserver/routers/common_api_types.py
poornimaramesh: Thanks! I've added a 0. default to this for now. I've noted that this should be another thing to take care of in the ticket from the previous comment

src/xngin/apiserver/routers/experiments/experiments_common.py
poornimaramesh: I explicitly want the denominator to be non-null outcomes, not all assignments.  In the last bandit experiment use case, that's the fraction I found myself computing offline a lot (i.e. of the users for whom we observed outcomes, how many actually completed onboarding).

Would you recommend switching the denominator to all assignments anyway?

src/xngin/apiserver/routers/common_api_types.py
poornimaramesh: Good catch AI-powered Doug, thanks :)
It looks like we're actually missing a critical validation step at experiment creation time for the DWH MABs -- we should be checking that the reward-type matches the `DataType` of the `target_field_name`, so I've added that now. That + the autofail value vs reward type check now cover this edge case also.

src/xngin/apiserver/routers/experiments/experiments_common.py
dctaf: That makes sense, but I think the migration leaves us with some rows that don't satisfy implied invariants -- outcome and autofailed_outcome should probably either both be null, or both be non-null, otherwise we have a 4th nonsensical state, I think? 

The potential for SQL+NULL footguns caught my eye, but we should double-check how this interacts with running analysis on existing/ongoing/historical experiments.

My understanding is that there are now three states we want to represent, using two fields:

1. no outcome: outcome == null, autofailed_outcome == null
2. autofailed outcome: outcome == X, autofailed_outcome == True
3. reported outcome: outcome == X, autofailed_outcome == False

The only valid state is both-null or both-non-null. But the migration only adds `autofailed_outcome`, which will be NULL. This leaves the potential for Draw rows to have a non-null `outcome` but a null `autofailed_outcome`.

This will only affect experiments that existed before this feature was implemented, but it is on the analysis path so it could generate misleading analysis results.

This might be fixed with an additional manual migration:

```
  UPDATE draws
  SET autofailed_outcome = false
  WHERE outcome IS NOT NULL;
```

Also, we could add a CHECK constraint (https://docs.sqlalchemy.org/en/20/core/constraints.html#check-constraint) to ensure this doesn't get mixed up in the future.

dctaf (changes requested):

src/xngin/apiserver/routers/experiments/experiments_common.py
poornimaramesh: That's a good point, thanks! I've added a check constraint and backfill for autofail_outcome values to set it to False.

dctaf (approved):

PR #310, opened 2026-08-07
Resolve most open dependabot issues.

dctaf:
This should resolve most of the open dependabot issues. Changelogs were
reviewed. This should not cause any undesirable behavioral changes.

I'll send a followup to replace python-jose with PyJWT because python-jose is not as actively maintained.


poornimaramesh (approved):
lgtm, approving, thanks!

PR #311, opened 2026-08-07
Replace python-jose with PyJWT

dctaf:
Replaces python-jose with PyJWT because the latter is actively maintained.

One behavioral change: the JWT iat/nbf claims are now verified w/a 30s leeway.

- **Pin the behavior of the Google ID token validation methods.**
- **Replace python-jose with PyJWT.**
- **Bump to cryptography 50.0.0.**
- **Drop the auth_callback endpoint test.**
- **Remove misleading comment from pyproject.toml.**


poornimaramesh (approved):
Looks good to me, thanks!

PR #312, opened 2026-08-11
Phase 3 of DWH-MABs: Use Snapshotter cadence to actually read outcomes 

markbotterill:
[DWH Native bandits Jira
](https://idinsight.atlassian.net/browse/EVE-31)
## Related PRs

Frontend: WIP! (wizard copy, integration-guide callout, snapshot Details column)

## Description

MAB-DWH experiments now read their outcomes from the org's data warehouse instead of having them pushed to our API.

- `ingest_dwh_outcomes()` selects draws with no outcome, fetches the target column for those participants via `get_participant_metrics` (the same helper A/B analysis uses, over the same connection, so no new access), and applies each value through the existing per-outcome bandit update. Pull and push therefore share one update path.
- It opens its own session and commits per outcome, so an interrupted run keeps what it applied and the next run resumes from the remaining NULL-outcome draws. Rejected and NULL values stay NULL, so a corrected warehouse value heals on a later run.
- Each snapshot for this type ingests before recording the analysis, and reports `ingested/invalid/pending` on the snapshot message.
- `POST .../outcome` returns a `LateValidationError` for `mab_online_dwh`, so outcomes have exactly one source. Sample calls drop the outcome example for the type.
- `--interval` becomes `--max-snapshot-age`, and `--mab-dwh-max-snapshot-age` (default 1 hour).

No new tables, columns or migrations. Behaviour for every other experiment type is unchanged.

## Future Tasks (optional)

- [EVE-165](https://idinsight.atlassian.net/browse/EVE-165) — decouple `Snapshot.data` from `ExperimentAnalysisResponse`.
- [EVE-166](https://idinsight.atlassian.net/browse/EVE-166) — adopt push/pull as standard vocabulary across API docs and product copy.
  - `start_date`/`end_date` are editable, so shortening `end_date` can strand pending outcomes. @qixotic's feedback would be useful here

  ## How has this been tested?

All tests + new ones pass

  ## To-do before merge (optional)

- Finalize `--mab-dwh-max-snapshot-age` default: 1hr right now w/ one minute of margin, so a slow run plus could drop ingestion to 2-hourly. 
- Raise the frontend PR.

  ## Checklist

  Fill with `x` for completed.


src/xngin/apiserver/routers/experiments/experiments_common.py
```diff
@@ -1157,6 +1155,117 @@ async def update_bandit_arm_with_outcome_impl(
     return arm_to_update
 
 
+@dataclass
```
dctaf: Default to using `@dataclass(slots=True)` unless there is some reason to avoid slots (there is usually not).

src/xngin/apiserver/routers/experiments/experiments_common.py
```diff
@@ -1157,6 +1155,117 @@ async def update_bandit_arm_with_outcome_impl(
...
+                    )
+                    .order_by(tables.Draw.created_at)
+                )
+            ).scalars()
+        )
+        if not pending_ids:
+            return report
+
+        dsconfig = experiment.datasource.get_config()
+        async with DwhSession(dsconfig.dwh) as dwh:
```
dctaf: Can we restructure this so that we aren't holding both the per-experiment xngin database connection (line 1191)and the customer DWH connection open at the same time (line 1220)? Avoiding simultaneous connections will make it easier to debug and design.

src/xngin/apiserver/routers/experiments/experiments_common.py
```diff
@@ -1157,6 +1155,117 @@ async def update_bandit_arm_with_outcome_impl(
...
+    """Read newly landed outcomes for a MAB_ONLINE_DWH experiment from the org's DWH and apply
+    them through the standard per-outcome bandit update.
+
+    Opens its own database session so that each outcome commits independently of any transaction
+    held by the caller.
+
+    An interrupted run keeps the outcomes already committed; the next run
+    picks up the remainder because ingestion selects only draws that still have no outcome.
+    """
+    report = IngestReport()
```
dctaf: Nit: move `report = IngestReport()` to be closer to where it is first used.

src/xngin/apiserver/snapshots/snapshotter.py
```diff
@@ -22,24 +22,44 @@
 # The snapshotter cron job can specify a different timeout via command line flags.
 SNAPSHOT_TIMEOUT_SECS = 90
 
+# Two independent things decide when an experiment gets a new snapshot:
+#   1. How often the cron wakes up and looks. Set in railway.snapshots.json, currently hourly.
+#   2. How old an experiment's newest snapshot must be to qualify when it does look. That is the
+#      age limit below, and it is a staleness limit at the default of
+#      6 hours against an hourly cron, an experiment qualifies on roughly one run in six.
```
dctaf: Bullet 2 can be reworded for clarity. Also, to avoid duplicating the informative comments here and in the Typer CLI help, and to keep them consistent, consider using the Typer CLI help as the single source of truth about this value.

src/xngin/apiserver/snapshots/snapshotter.py
```diff
@@ -22,24 +22,44 @@
...
+DEFAULT_MAX_SNAPSHOT_AGE_SECS = 6 * 60 * 60
 
-async def create_pending_snapshots(snapshot_interval: int):
+# MAB-DWH is the one type whose snapshot also ingests outcomes from the org's data warehouse, so
+# for it this age limit doubles as how often we pull. Kept at or below the cron period so that
+# every run ingests. Applies to MAB-DWH only; plain MABs, CMABs and all frequentist types use the
+# default above.
+DEFAULT_MAB_DWH_MAX_SNAPSHOT_AGE_SECS = 60 * 60
```
dctaf: 1. Consider replacing these multiplication expressions with `timedelta(minutes=60).seconds`. That might make it easier to read.

2. I'm wondering if this actually runs ingestion hourly, or less frequently than hourly. The updated_at columns are set to the time the row is updated, not to when we start the snapshot, and the +1 minute offset might lead to skips. My understanding is that we do want ingestion to happen every hour, whereas we only need snapshots to run once approximately every six hours, so the requirements are subtly different here. Also, customers may want to increase sampling frequency or decrease it, and we might want to allow customers to adjust analysis freq separately from pull frequency. Let's discuss this in our 1:1.

src/xngin/apiserver/snapshots/snapshotter.py
```diff
@@ -141,16 +170,29 @@ async def _handle_one_snapshot_safely(session: AsyncSession, snapshot: tables.Sn
...
         try:
             async with asyncio.timeout(snapshot_timeout):
+                if ExperimentsType(experiment.experiment_type) == ExperimentsType.MAB_ONLINE_DWH:
+                    # Ingestion commits per outcome on its own session; refresh this session's
+                    # arms so the analysis below sees the updated posteriors.
+                    ingest_report = await experiments_common.ingest_dwh_outcomes(experiment.id)
+                    ingest_message = ingest_report.summary()
+                    logger.info(f"{experiment.id}.{snapshot.id}: {ingest_message}")
+                    for arm in experiment.arms:
+                        await session.refresh(arm)
```
dctaf: These refreshes are workarounds for SQLAlchemy's in-memory cache, and refer to the arms directly despite their invalidation being a side effect of the ingestion/update. We can't really remove the need for it without a larger refactor, but I believe we can do a single `await session.refresh(experiment)` instead of looping over the arms, which would be safer against future code changes.

src/xngin/apiserver/snapshots/snapshotter.py
```diff
@@ -141,16 +170,29 @@ async def _handle_one_snapshot_safely(session: AsyncSession, snapshot: tables.Sn
...
                 result = await _query_dwh_for_snapshot_data(session, datasource, experiment)
                 snapshot.data = result.model_dump(mode="json")
                 snapshot.status = "success"
+                snapshot.message = ingest_message
```
dctaf, line 179: Would it be useful to track ingestion message vs analysis snapshot message separately? More structure here could make it easier to monitor. Consider adding additional fields to the Snapshot table.

src/xngin/apiserver/snapshots/snapshotter.py
```diff
@@ -51,12 +71,21 @@ async def create_pending_snapshots(snapshot_interval: int):
             .order_by(tables.Experiment.id, tables.Snapshot.updated_at.desc())
             .cte()
         )
+
+        # How old an experiment's newest snapshot must be before it earns another one.
+        max_snapshot_age = case(
+            (
+                experiments_snapshot_status.c.experiment_type == ExperimentsType.MAB_ONLINE_DWH.value,
+                text(f"interval '{mab_dwh_max_snapshot_age_secs} seconds'"),
```
dctaf, line 71: In this architecture, `mab_dwh_max_snapshot_age_secs` probably needs to be less than the scheduled cron interval if we want this to run hourly.

dctaf (changes requested):

src/xngin/apiserver/routers/experiments/experiments_common.py
markbotterill: TIL about `slots`! Thank you - implemented 🙏

src/xngin/apiserver/routers/experiments/experiments_common.py
markbotterill: Agreed - I've sorted it into 3 clear phases to avoid the overlaps / nested connections.

src/xngin/apiserver/routers/experiments/experiments_common.py
markbotterill: Sorted 🫡

src/xngin/apiserver/snapshots/snapshotter.py
markbotterill: Shifted to Typer comments!

src/xngin/apiserver/snapshots/snapshotter.py
markbotterill: this is sorted!

src/xngin/apiserver/snapshots/snapshotter.py
markbotterill: I've handled 1. 2 I am parking until we chat!

src/xngin/apiserver/snapshots/snapshotter.py
markbotterill: I think this touches on the debate above around whether we are stretching the definition of a snapshot to be this large umbrella that also really means just doing a DWH-data-pull. If we agree on splitting the schedule for these two tasks, maybe it makes sense to consider them different entities (or perhaps we can spec them out as different flavours of Snapshot?)

PR #315, opened 2026-08-15
Refine analyze_experiment implementation and fix some pandas typing.

dctaf:
IntelliJ noted that our pandas-stubs dependency did not match the pinned pandas
version. Upgrading the stubs led to mypy failing, revealing that we were using
a Pandas expression that the stubs couldn't type correctly. I rewrote the
expression so that it passes type checking. In the process, this reduces memory
usage about 50% (for < 10 metrics) and improves performance. Additional pinning
tests were added to ensure we don't see behavioral changes.


markbotterill (approved):
LGTM 🚀 V nice that we get some performance gains here too!

PR #316, opened 2026-08-20
Migrate tests of bandit outcome update to API layer.

dctaf:
## Ticket

EVE-180

## Description

Moves some of the tests of the bandit outcome update behavior to the API layer.

This is helpful because the SQLAlchemy ORM has some surprising behavior that
can make tests difficult to implement. Testing at the API layer ensures we take
those out of consideration when testing, and helps us ensure we are observing
customer-visible effects.

We also strengthen some of the tests based on observed behavior.


src/xngin/apiserver/routers/admin/test_admin_api.py
```diff
@@ -2572,52 +2572,6 @@ def test_cmab_experiments_analyze(testing_bandit_experiment, aclient: AdminAPICl
         assert analysis.post_pred_stdev is not None
 
 
-async def test_create_and_update_outcome_mab_dwh_happy_path(
```
poornimaramesh, line 2575: we've removed this test, but don't replicate it fully anywhere else.
Specifically, with this test gone, we don't test the create and get MAB_ONLINE_DWH experiment path at all. Maybe we add tests for those actions here? It looks like get and update assignment are covered by the other tests you've added though.

src/xngin/apiserver/routers/experiments/test_experiments_api.py
```diff
@@ -1153,78 +1177,434 @@ async def test_get_assignment_mab_cache_headers(
...
+        eclient.update_bandit_arm_with_participant_outcome(
+            api_key=testing_datasource.key,
+            body=UpdateBanditArmOutcomeRequest(outcome=1.0),
+            experiment_id=experiment_id,
+            participant_id=participant_id,
+        )
+    assert exc.value.result.status == HTTPStatus.UNPROCESSABLE_CONTENT
+
+
+async def test_create_mab_dwh_bool_target_with_normal_reward_returns_422(testing_datasource, aclient: AdminAPIClient):
```
poornimaramesh: shouldn't this be in `test_admin_api`? -- this is a test that blocks experiment creation, which is an admin action...

src/xngin/apiserver/routers/experiments/test_experiments_api.py
```diff
@@ -1153,78 +1177,434 @@ async def test_get_assignment_mab_cache_headers(
...
+            api_key=testing_datasource.key,
+            body=UpdateBanditArmOutcomeRequest(outcome=42.7),
+            experiment_id=experiment.experiment_id,
+            participant_id=assignment.participant_id,
+        )
+    assert exc.value.result.status == HTTPStatus.UNPROCESSABLE_CONTENT
+    assert "Cannot dynamically update arms for frequentist experiments" in str(exc.value.result.data)
+
+
+@pytest.mark.skip("EVE-171")
```
poornimaramesh, line 1522: if this is meant specifically to be testing the multiple outcome problem, should we remove the skip? Or remove the test when we fix the problem?

src/xngin/apiserver/routers/experiments/test_experiments_api.py
```diff
@@ -1153,78 +1177,434 @@ async def test_get_assignment_mab_cache_headers(
...
+
+    Normal/Bernoulli is the only update_arm branch that consumes every entry in its outcomes
+    argument; the Beta/Bernoulli and Normal/Normal branches use only outcomes[0]. This makes it the
+    branch that detects if the endpoint accidentally supplies duplicate or previously absorbed
+    outcomes instead of the intended singleton. Draw rows continue to retain the complete outcome
+    and context history, so a future batch implementation can deliberately recompute from the
+    original prior without weakening this incremental-update invariant.
+    """
+    # A single arm dimension keeps the posterior easy to read. Arms start at
+    # mu=[mu_init] and covariance=diag([sigma_init]) (storage_format_converters.py:532).
```
poornimaramesh: if we're keeping ths test, should we remove the comments?

src/xngin/apiserver/routers/experiments/test_experiments_api.py
```diff
@@ -1153,78 +1177,434 @@ async def test_get_assignment_mab_cache_headers(
...
+    updated_arm = eclient.update_bandit_arm_with_participant_outcome(
         api_key=testing_datasource.key,
         body=UpdateBanditArmOutcomeRequest(outcome=1.0),
-        experiment_id=cmab_experiment.experiment_id,
+        experiment_id=experiment_id,
         participant_id="1",
+    ).data
+
+    # Oracle: the same model fitted with the one outcome that was actually recorded. Using
+    # update_arm itself keeps the expectation free of any reimplementation of the math.
```
poornimaramesh: same comment as above. I am not sure what this means either -- `update_arm` explicitly implements the update math, so we don't replicate it anywhere else in the code or in the tests

src/xngin/apiserver/routers/experiments/test_experiments_api.py
```diff
@@ -1153,78 +1177,434 @@ async def test_get_assignment_mab_cache_headers(
...
-async def test_update_bandit_arm_with_outcome_rejects_non_numeric_outcome(
-    testing_datasource, aclient: AdminAPIClient, eclient: ExperimentsAPIClient
+@pytest.mark.parametrize(
+    ("experiment_type", "prior_type", "reward_type"),
+    [
+        (ExperimentsType.MAB_ONLINE, PriorTypes.BETA, LikelihoodTypes.BERNOULLI),
+        (ExperimentsType.MAB_ONLINE, PriorTypes.NORMAL, LikelihoodTypes.NORMAL),
+        (ExperimentsType.MAB_ONLINE, PriorTypes.NORMAL, LikelihoodTypes.BERNOULLI),
+        (ExperimentsType.CMAB_ONLINE, PriorTypes.NORMAL, LikelihoodTypes.NORMAL),
+        (ExperimentsType.CMAB_ONLINE, PriorTypes.NORMAL, LikelihoodTypes.BERNOULLI),
```
poornimaramesh, line 1209: could we add Experiments.MAB_ONLINE_DWH with the same combination of prior and likelihood as for MAB_ONLINE to the parameters? We have no happy path tests left for this experiment type right now, since we've deleted them from test_experiments_common

poornimaramesh (changes requested):
Looks good overall -- the rearranging of the tests makes sense in general. Left some comments to add / cover test paths that we've now dropped because of the move.

src/xngin/apiserver/routers/admin/test_admin_api.py
dctaf: Restored.

src/xngin/apiserver/routers/experiments/test_experiments_api.py
dctaf: Done.

src/xngin/apiserver/routers/experiments/test_experiments_api.py
dctaf: Done

src/xngin/apiserver/routers/experiments/test_experiments_api.py
dctaf: Let's remove the test when we fix the bug.

src/xngin/apiserver/routers/experiments/test_experiments_api.py
dctaf: Removed.

src/xngin/apiserver/routers/experiments/test_experiments_api.py
dctaf: Removed.

poornimaramesh (approved):
Looks good, thanks!

PR #317, opened 2026-08-20
Move responsibility for managing bandit update transaction to callers.

dctaf:
## Ticket

EVE-180

## Description

Instead of having transactions committed or operations flushed per-bandit
update, we moves responsibility for calling .commit()/.flush() to the API
endpoint and autofailer.

This paves the way to refactor the bandit updater to be invoked multiple times
in a single database transaction. Also, it is generally helpful to couple the
API implementation to the transaction lifecycle, moreso than it is to have deep
methods run .commit(), because the latter increases the chance that we might
experience an error between the .commit() and when we compose the API response,
leading to misleading representation of our state.

We also replace some ORM mutations with a direct update() call. This is
equivalent behavior today, but later we can defer the arm updates to minimize
contention.

### DIFF
--- a/src/xngin/apiserver/snapshots/test_autofail.py
+++ b/src/xngin/apiserver/snapshots/test_autofail.py
@@ -459,6 +459,7 @@ async def test_process_pending_autofail_updates_survives_a_failure(
 
     async def fail_first_participant(*args, **kwargs):
         if kwargs["participant_id"] == "0":
+            await original(*args, **kwargs)
             raise RuntimeError("boom")
         return await original(*args, **kwargs)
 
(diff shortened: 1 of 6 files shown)


src/xngin/apiserver/snapshots/test_autofail.py
```diff
@@ -468,6 +469,12 @@ async def fail_first_participant(*args, **kwargs):
 
     updates = await get_autofail_updates(xngin_session)
     assert [(u.participant_id, u.status) for u in updates] == [("0", "failed"), ("1", "success")]
+    draws = (
+        await xngin_session.execute(
+            select(tables.Draw).where(tables.Draw.experiment_id == experiment_id).order_by(tables.Draw.participant_id)
+        )
+    ).scalars()
+    assert [(draw.participant_id, draw.outcome) for draw in draws] == [("0", None), ("1", 0.0)]
```
poornimaramesh, line 477: we don't have to worry about it in this PR, but it is problematic if we lose an update for a particular participant with the autofail feature (particularly if it's buried somewhere in the middle of a flurry of updates) -- we don't have any retries, and we lose valuable information that determines how fast a bandit experiment will converge (!). We should probably build out something on the UI that surfaces this error for the user, allows retries for failed updates and/or re-queues failed updates. Made a ticket to track this [here](https://idinsight.atlassian.net/browse/EVE-189)

poornimaramesh (approved):
Looks good overall, thanks!

PR #318, opened 2026-08-20
Refactor autofailer to remove use of queue table.

dctaf:
## Ticket

EVE-180

## Description

This replaces the autofailer's autofail_updates table with a Postgres partial
index and some queries.

The design of the existing autofailer was inspired by the snapshotter. The
snapshotter has different operational requirements than the autofailer. Unlike
snapshots, autofail does not talk to customer DWH, so retries and timeouts and
reporting are less important. Another way it is different is that the
autofailer mutates shared state, whereas the snapshotter creates new rows on
every run. The autofailer is probably CPU bound, whereas the snapshotter is I/O
bound.

We implement the new autofailer using a partial index (to efficiently identify
candidate draws), using a single transaction for multiple draws instead of one
transaction per draw, and processes draws in bulk. It is tolerant to business
logic bugs preventing a single experiment from autofailing, and also has a
nod to fairness so that one very large experiment doesn't cause problems for
smaller experiments.

This is not a perfect scheduler, but it should be adequate base to build on as 
EVE-170, 171, 181, 182, 183, etc are implemented.


src/xngin/apiserver/snapshots/autofail.py
```diff
@@ -16,143 +19,213 @@
...
+        .exists()
+    )
+    return (
+        select(tables.Experiment.id)
+        .where(
+            tables.Experiment.enable_autofail.is_(True),
+            tables.Experiment.experiment_type.in_(AUTOFAIL_EXPERIMENT_TYPES),
+            has_pending_draw,
+        )
+        .order_by(func.random())  # return experiment IDs in a random order
```
poornimaramesh, line 69: do we really need the `order_by` particularly if they are going to be randomly ordered?

poornimaramesh: From looking through the double-ended queue logic below, this is maybe a bigger question to discuss: in the case where we expect the autofail job to fail for a subset of draws, do we prefer updating as many experiments as possible and spreading the failures _across_ experiments, or do we prefer keeping the failures contained to a small subset of experiments? I would argue that the latter would be better i.e. if we do expect failures, we try and keep them contained to a few experiments (especially if we expect to surface these failures to the user, and maybe enable manual re-queuing). However, from an engineering point of view, I guess that would lock us into a specific sequence of updates for experiments, risk having failures always occur due to timeout or something else at the end of the queue, and therefore we systematically end up preferring some experiments for successful autofail updates over others.

I am ok with leaving things as is for now, and creating a tracking ticket for us to revisit this if it becomes an issue, but happy to discuss if there are other ways to approach this!

poornimaramesh (approved):
Looks good overall, approving, thanks for the updates!

I agree that retries, timeouts and reporting aren't nescessary right now, because not talking to a customer dwh makes them likely to fail. However, we should think about surfacing them to a user anyway, since failures will affect how quickly the experiment converges, even if we switch to a different algorithm for batched updates. Silently (for the user) losing a whole batch of updates without a fallback in place would be pretty bad for decision-making / running an experiment even if they get requeued the next time an autofail job runs.

PR #321, opened 2026-08-26
Batch cluster ICC/CV lookups into a single DWH query for power checks

mdulieu:
## Description

Power checks on cluster-randomized designs issued 2 DWH queries per metric: one for cluster-size stats (identical result every time, since it never reads the metric column) and one for ICC data. A 4-metric check meant 8 sequential round trips to the customer's warehouse.

This PR fetches the cluster key and all outcome columns in **one query** and computes the stats in pandas:

- `get_cluster_outcome_data` now takes a list of outcome columns and no longer filters nulls per metric in SQL; `get_cluster_size_stats` is deleted (its aggregates are now derived from the same fetched rows).
- `calculate_icc_and_cv_from_database` is renamed to `calculate_cluster_stats_from_database`: cluster sizes/CV are computed once over the full frame (nulls included, since metrics are outcomes that may be filled in as the experiment runs), ICC per metric on that metric's non-null rows, with `ddof=0` to match SQL `stddev_pop`.
- `power_check` makes one batched call for the metrics without user-supplied ICC (and none if all are user-supplied).

Behavior-preserving: no API changes, same numbers, same error messages/types. 

## Testing

- All existing power-check API tests pass with their exact asserted values unchanged (incl. the null-cluster-key test pinning ICC 0.021 / avg size 13.5 / CV 0.282).
- New: `test_batched_matches_individual_queries` (batched calls match per-metric calls, covers the BOOLEAN cast) and `test_null_outcomes_dropped_per_metric_but_counted_in_cluster_sizes` (nulls counted in cluster sizes, dropped from ICC).
- `test_power_adapters.py` migrated to the new signature and to module-level test functions per the style guide; asserted values untouched.
- Full admin suite 129/129, dwh+stats 318/318, mypy and ruff clean.


src/xngin/apiserver/routers/power_adapters.py
```diff
@@ -1,49 +1,72 @@
...
+    for outcome_column in outcome_columns:
+        # ICC uses only rows with a value for this outcome, mirroring the per-metric
+        # "outcome IS NOT NULL" filter used when each metric was queried separately.
+        outcome_df = df[[cluster_column, outcome_column]].dropna(subset=[outcome_column])
+        if outcome_df.empty:
+            raise LateValidationError(
+                f"No data found for cluster column '{cluster_column}' and outcome '{outcome_column}'"
+            )
+        try:
+            icc = calculate_icc_from_dataframe(outcome_df, cluster_column=cluster_column, outcome_column=outcome_column)
```
dctaf, line 63: One thing to consider as Evidential scales is that the # of participants in the experiments could get very large. One thought experiment I like to use is "What happens if there are 1M participants?".

Because calculate_icc_from_dataframe is pure, it might be possible to replace this with SQL, which might improve this implementation by reducing the amount of data transferred, computation, and also reduce memory pressure. There are tradeoffs to consider, but the general shape would be something like this:

```postgresql
SELECT
  school_id,
  count(*)                                       AS n_rows,

  count(test_score)                              AS test_score_n,
  sum(test_score)                 AS test_score_sum,
  sum(test_score * test_score)               AS test_score_sumsq,

  count(passed)                                  AS passed_n,
  sum(...)    AS passed_sum,
  sum(... * ...)  AS passed_sumsq
FROM dwh
WHERE school_id IS NOT NULL
  AND <filters from create_query_filters>
GROUP BY school_id
```

The remaining stats can probably be derived from that single row. This might require some backend-specific variation BigQuery-specific implementation. I don't think this is important to do for this PR but it is something to consider!

src/xngin/apiserver/routers/admin/admin_api.py
```diff
@@ -2280,6 +2280,24 @@ async def power_check(
...
+                    dwh.session,
+                    sa_table,
+                    cluster_key,
+                    db_derived_metrics,
+                    filters,
+                )
+                if db_derived_metrics
+                else {}
+            )
             for metric_stat in metric_stats:
```
dctaf, line 2305: Minor refactoring opportunity: generally, we should decouple these two functions:

1. Querying the customer data warehouse,
2. Computing on the results of that query.

1 only happens while the DwhSession context manager is active. 2 can happen after that block (and the remote database connection) is closed.

Keeping that distinction in mind is helpful for a few reasons. One is that it ensures we have a clean inventory of the queries we are running, which will make it easier for us to add new data sources in the future. Another is that it lets us make individual methods easier to test because we can fake their inputs more easily. Another is that it will make exceptions easier to track down because any error we diagnose within the DwhSession() block requires us to consider whether there are threading or remote database issues in play. And another benefit is that it makes it very easy to determine which sections of code talk to the DWH so that we can keep it as short as possible. Here's an oversimplified idealized example:

```python
with DwhSession(...) as dwh:
  metric_stats = await dwh.get_stats_on_metrics(...)
  cluster_stats = await dwh.get_cluster_stats(...) if cluster_key is not None
derived_metric_stats = math_on_metric_stats(metric_stats)
cluster_stats = math_on_cluster_stats(cluster_stats) if cluster_stats is not None
```

That's a refactor that can happen in a later PR but I mention it here because I want us to keep this in mind.

dctaf (approved):
The comments pertain to future opportunities. This PR looks good. Thanks!

dctaf:
Adding @qixotic as a reviewer for final approval and merge.

mdulieu:
Thank you for the comments, I'll keep those in mind for future PRs.

qixotic:
Verified test_queries.py passes against our bq:

```
DATABASE_URL="postgresql+psycopg://postgres:postgres@localhost:5499/test_xngin?sslmode=disable" \
XNGIN_QUERIES_TEST_URI="bigquery://..." \
uv run pytest -v -rA src/xngin/apiserver/dwh/test_queries.py
```

PR #322, opened 2026-08-26
Honor desired_n_clusters in power checks

mdulieu:
## Summary

The power endpoint previously only accepted a desired sample size in individuals (`desired_n`), so for cluster-randomized designs the frontend converted clusters to individuals itself using a single `avg_cluster_size` from a prior response. This PR makes the power check honor `desired_n_clusters` (which already exists on `PreassignedFrequentistExperimentSpec` for preassigned sampling) directly:

- `check_power` accepts `desired_n_clusters` and converts it per metric as `round(desired_n_clusters * metric.avg_cluster_size)`. When both fields are set, `desired_n_clusters` takes precedence over `desired_n`. A metric without `avg_cluster_size` raises a clear `StatsPowerError` (unreachable via the endpoint, which fills cluster stats whenever `cluster_key` is set).
- The power endpoint passes `desired_n_clusters` from the design spec through to `check_power`. The MDE lands in the existing `pct_change_with_desired_n` response field.
- No request/response schema changes; OpenAPI descriptions of `desired_n`, `desired_n_clusters`, and `pct_change_with_desired_n` updated to document the semantics.

## Testing

- Unit tests: per-metric conversion (two metrics differing only in `avg_cluster_size`), precedence when both fields are set, and the missing-cluster-stats error.
- Endpoint test: 40 clusters × avg_cluster_size 10 returns the identical MDE as `desired_n=400`, and precedence holds end-to-end.
- No frontend change needed: the FE already sends `desired_n_clusters` in power requests for cluster designs, so this change takes effect on deploy (the FE-converted `desired_n` it also sends is superseded per the precedence above). Client regen was verified to produce no diff — the field already existed in the schema and this PR only changes descriptions. One visible effect: the Estimated MDE shown for cluster designs may shift slightly, since the MDE is now computed from `desired_n_clusters × per-metric avg_cluster_size` server-side instead of the frontend's single client-side conversion.


src/xngin/stats/power.py
```diff
@@ -123,33 +124,45 @@ def check_power(
...
     """
     analyses = []
     for metric in metrics:
+        metric_desired_n = desired_n
+        if desired_n_clusters is not None:
+            if metric.avg_cluster_size is None:
+                raise StatsPowerError(
+                    f"desired_n_clusters requires cluster statistics, but metric {metric.field_name} "
+                    "has no avg_cluster_size."
+                )
```
poornimaramesh, line 151: if desired_n_clusters requires avg_cluster_size to be set, could we write an explicit check in the pydantic schema instead?

src/xngin/apiserver/routers/common_api_types.py
```diff
@@ -1114,7 +1117,9 @@ class PreassignedFrequentistExperimentSpec(BaseFrequentistDesignSpec):
             ge=1,
             description=(
                 "Desired number of clusters to sample when creating a cluster-randomized preassigned experiment. "
-                "Only valid when cluster_key is set. All eligible participants in each sampled cluster are included."
+                "Only valid when cluster_key is set. All eligible participants in each sampled cluster are included. "
+                "In power checks, the minimum detectable effect is computed for this many clusters, converted to a "
+                "per-metric sample size using each metric's avg_cluster_size; takes precedence over desired_n."
             ),
         ),
     ] = None
```
poornimaramesh, line 1125: see comment about adding an explicit check in the `validate_cluster_randomization` check so avg_cluster_size is set when desired_n_clusters is set

poornimaramesh (approved):
Added a minor comment about where to catch the desred_n_clusters and avg_cluster_size error. Looks good otherwise, thanks!

src/xngin/stats/power.py
mdulieu: A schema check would break the common case where only desired_n_clusters is set and avg_cluster_size isn't in the request at all: the endpoint computes it from the DWH (per metric) after validation, so it doesn't exist yet when the schema is checked. This check runs after that step, where the value is guaranteed to exist unless check_power was called directly (e.g. from tests).

src/xngin/apiserver/routers/common_api_types.py
mdulieu: Answered on the power.py comment. The short version is that avg_cluster_size often isn't present in the request; it's derived from the DWH after validation.

PR #323, opened 2026-08-26
fix precision errors for bandit weights to normal prior params approximation

poornimaramesh:
## Ticket

[EVE 186](https://idinsight.atlassian.net/browse/EVE-186)

## Description
Updated the conversion for weights -> priors for the normal distribution -- the error reported in the ticket was because we were running into precision errors (i.e. we round up to 2 decimal places, and some times the μ values are only significant from the 3rd decimal place onwards for higher context numbers).

I use a very hacky approximation that will break down as with increasing contexts for the CMABs. However it should hold for small numbers of users (<10k users) before the weights start to get updated.

We should revisit this if/when we actually have people using the CMAB experiments -- tracking ticket [here](https://idinsight.atlassian.net/browse/EVE-191?atlOrigin=eyJpIjoiMThjNjRhODdhYjNmNDA5MGIwNmIwYzg5NTUxNGE2YjgiLCJwIjoiaiJ9)

## How has this been tested?
Check that CMABs with ~10 contexts and arm weights [10, 20, 30, 40] return different μ values instead of zeros from the frontend. Run updated tests.

## Checklist


src/xngin/stats/bandit_weights_to_prior.py
```diff
@@ -78,16 +78,21 @@ def objective(params: np.ndarray) -> float:
...
     """
     Convert bandit weights to Normal prior parameters (mu, sigma) for each arm.
 
+    For multi-dimensional Normal distributions (CMABs), the mean parameters are optimized
+    to minimize the squared error between the expected probabilities and the probabilities derived
+    from the univariate Normal cdf -- this is a simplification in order to avoid precision errors
+    from computing the multivariate Normal cdf.
+    As the number of dimensions increases, the approximation diverges from the true probabilities.
+    However, this is a reasonable error tolerance for the purposes of setting prior parameters for CMABs.
+
```
dctaf, line 91: Nit: stray empty line

dctaf (approved):

PR #324, opened 2026-08-26
Canonicalize representation of NULLs in JSONB columns.

dctaf:
Ticket: EVE-192

SQLAlchemy's default behavior when assigning None to a nullable JSONB column is
to store JSON `null`, not a Postgres NULL. When the Postgres column allows
NULLs, this allows a JSONB null to be interpreted differently from a Postgres
NULL. For example, a SQLAlchemy query of .is_(None) would match only Postgres
NULLs, not JSONB's `null`. Inserts of rows where the field is not assigned to
would get a Postgres NULL. This led to some columns having a mix of Postgres
NULL and JSONB null, without any Python-visible difference between the two,
because Python's `x is None` evaluates to True for JSONB `null` or Postgres
NULL. 

This PR adds `none_as_null=True` to all current nullable JSONB columns. This
changes the assignment behavior: `table.x = None` now writes a Postgres NULL,
not a JSONB null. We also rewrite any production JSONB nulls to Postgres NULL
via SQL migrations.

The migration query will affect rows in `experiments.{power_analyses,balance_check,table_list}` 
and `turn_connections.journeys_dict`.


poornimaramesh (approved):

PR #326, opened 2026-08-28
fix precision errors; fix overdeterminism

poornimaramesh:
## Ticket

Fixes [EVE 186](https://idinsight.atlassian.net/browse/EVE-196)

## Description
The check for equality of the input weights and clamping the parameter values to prevent overdeterminism in the root-finding algorithm were both leading to precision errors with increasing number of arms.

Implemented fixes: updated the equality check with a relative instead of absolute tolerance; updated the root-finding algorithms so we don't clamp the parameters -- we choose to live with the overdeterminism.


src/xngin/stats/bandit_weights_to_prior.py
```diff
@@ -23,6 +21,10 @@ def bandit_weights_to_beta_prior(
             parameters for the Beta distribution.
     """
     expected_probabilities = np.asarray(expected_probabilities, dtype=np.float64)
+
+    if expected_probabilities.sum() != 100:
```
dctaf: Equality tests on floating point numbers are always worth a second look. There is a validator on BaseDesignSpec (src/xngin/apiserver/routers/common_api_types.py:848) using `math.isclose(total, 100.0, rel_tol=1e-9)` to determine if they sum to 100, and has a comment about how it is trying to match stochatreat's behavior. Using `!=` here will allow API validation to succeed but they could fail here, which will generate inconsistent client-side error message. Example inconsistent inputs: `[2.04, 89.81, 8.15]`. I don't have a suggestion for the best way to handle this, but it seems like something we should have a standard practice for.

This also applies on line 111.

src/xngin/stats/bandit_weights_to_prior.py
```diff
@@ -23,6 +21,10 @@ def bandit_weights_to_beta_prior(
...
+        raise ValueError("Expected probabilities must sum to 100.")
+
     expected_probabilities *= 0.01  # Normalize to sum to 1
```
dctaf: `expected_probabilities` is a non-primitive type so this modifies the arguments in the caller's scope. The caller may not expect that. If we're mutating the input parameters, take care to clone them. This also applies on 114.

dctaf (changes requested):

dctaf (approved):

PR #327, opened 2026-08-28
Reuse client-provided metric baseline stats in power checks, skipping the dwh when possible

mdulieu:
Implements the following task: when a power check request already carries the
baseline stats about its metrics (as returned by a previous power check), reuse them instead of
re-querying the data warehouse.

**Stacked on #322** — the first commit here belongs to that PR; only the last commit
(`a898e34b`) is under review. The diff collapses to just this change once #322 merges.

## API change

`DesignSpecMetricRequest` gains five optional fields mirroring `DesignSpecMetric`:
`metric_type`, `metric_baseline`, `metric_stddev`, `available_n`, `available_nonnull_n`.
This resolves the existing TODO in `common_api_types.py` about accepting stats as inputs so the
endpoint can act as a general power calculator.

- A validator enforces the stats are all-or-none per metric (`metric_stddev` only for NUMERIC
  metrics), so a calculation never mixes supplied and dwh-derived numbers for one metric.
- The change is backward compatible: requests without the new fields behave exactly as before.

## Endpoint behavior

`power_check` now partitions metrics:

- Metrics carrying baseline stats are used as-is; only the rest are queried
  (`get_stats_on_metrics` gets the narrowed list).
- For cluster designs, the existing per-metric rule is unchanged: user-supplied ICC/CV/avg
  cluster size are used when present, otherwise derived from the dwh.
- When nothing needs the warehouse, every metric has baseline stats, plus ICC for cluster
  designs, the `DwhSession` is never opened. Note this also skips validating the design spec
  against the participants table (intentional: it makes the endpoint usable as a pure
  calculator, and the caller supplied the numbers being computed on).

The caller is responsible for echoing stats that match the request's filters; stale or
mismatched stats produce a power result for the stats given, as with manual ICC today.

## Tests

- Model-level: all-or-none validation, stddev-requires-NUMERIC, faithful request→metric
  conversion.
- Endpoint-level: echoing stats back with a nonexistent `table_name` succeeds and returns
  analyses identical to the queried run (proving the dwh was untouched), for both individual
  and cluster (`desired_n_clusters`) designs; a mixed request (one metric with stats, one
  without) matches a fully-queried run.

## Follow-ups

- FE: send the stats from the initial power response back on desired-size re-checks
  (`convertToFrequentistDesignSpec`).
- Merge note: overlaps with #321 in `power_check`'s cluster block; whichever lands second
  needs a trivial rebase.

### DIFF
--- a/src/xngin/apiserver/routers/admin/admin_api.py
+++ b/src/xngin/apiserver/routers/admin/admin_api.py
@@ -2248,7 +2248,13 @@ async def power_check(
     user: Annotated[tables.User, Depends(require_user_from_token)],
     body: PowerRequest,
 ) -> PowerResponse:
-    """Performs a power check for the specified datasource."""
+    """Performs a power check for the specified datasource.
+
+    Metrics carrying baseline stats (e.g. echoed back from a prior power check response) are used
+    as-is rather than re-queried from the data warehouse. When no metric needs the warehouse — every
+    metric has baseline stats, and cluster stats for cluster-randomized designs — the warehouse is
+    not contacted at all, so the design spec is not validated against the participants table.
+    """
     design_spec = body.design_spec
     ds = await get_datasource_or_raise(session, user, datasource_id)
     if isinstance(ds.config, NoDwh):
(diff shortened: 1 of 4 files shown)


poornimaramesh:
I'm a bit confused by this one. If we want to make a standalone power_check endpoint independent of the datasource, why not separate it from the existing power_check endpoint?
Also, my understanding was that the check against the database in the current endpoint was an implicit guard against instances where we have existing stats but we change the cluster_key or the metric (even though we wipe the PowerAnalysisRequest schema on the frontend), and this change breaks that no? 
Maybe we should make that check explicit also?

qixotic:
> I'm a bit confused by this one. If we want to make a standalone power_check endpoint independent of the datasource, why not separate it from the existing power_check endpoint? 

@poornimaramesh  This is purely an optimization for existing power calc behavior, but you are right in that it would support a standalone power check behavior _if we wanted to_. There's no proposal right now to actually make a FE that does that, but I did want this change specifically to not have to requery for the exact same stats just because I changed my desired_n ([EVE-36](https://idinsight.atlassian.net/browse/EVE-36)). Good eye, thanks!

Although that is an interesting suggestion, to have a separate endpoint for accepting user-supplied stats as a cleaner distinction. I wonder if that could address some of your other concerns @mdulieu about users mixing some user inputs with dwh-derived on the cluster stats side.  If a user decides to edit any field, we could switch the FE to using this "manual" power check endpoint. If so I'd want to make it clear to the user though that any re-click of the power estimation button would _not_ draw from the dwh unless they cleared all stats first.

Also this suggestion is a little related to @Snehaaa18's [EVE-194](https://idinsight.atlassian.net/browse/EVE-194) refactoring of the top level PowerRequest model, if you'd like to also weigh in!

src/xngin/apiserver/routers/common_api_types.py
```diff
@@ -168,6 +170,33 @@ class DesignSpecMetricRequest(DesignSpecMetricBase):
         ),
     ] = None
 
+    # Optional baseline stats, mirroring the fields of DesignSpecMetric. When all are provided
+    # (metric_stddev only for NUMERIC metrics), the server skips the dwh stats query for this metric.
+    metric_type: Annotated[
```
qixotic: Is this lifting up all the fields from a DesignSpecMetric into DesignSpecMetricRequest? If so, maybe they could just go directly into DesignSpecMetricBase since both DesignSpecMetricRequest and DesignSpecMetric have the fields. Then the respective request/response objects could just do their path-specific validation on them as needed. If we're sure we don't need different validation on the request vs response side of things, would this motivate collapsing all 3 models down into one?

mdulieu:
@qixotic, I have lifted the five stats fields to DesignSpecMetricBase. I kept the two subclasses because their validation differ. A request must carry one of metric_pct_change/metric_target, while a response carries both.

src/xngin/apiserver/routers/common_api_types.py
```diff
@@ -176,6 +180,41 @@ def check_has_only_one_of_pct_change_or_target(self) -> Self:
...
+        never mixes supplied stats with dwh-derived ones for the same metric."""
+        stats_fields = (self.metric_type, self.metric_baseline, self.available_nonnull_n, self.available_n)
+        if any(f is not None for f in stats_fields) and any(f is None for f in stats_fields):
+            raise ValueError(
+                "metric_type, metric_baseline, available_nonnull_n, and available_n must all be set "
+                "together or all be None"
+            )
+        if self.metric_stddev is not None and self.metric_type is not MetricType.NUMERIC:
+            raise ValueError("metric_stddev may only be set for NUMERIC metrics")
+        return self
```
poornimaramesh: looks like we're running the same check here as from the DesignSpecMetric. Why not just make this class a child of DesignSpecMetric (or add the stddev check to `DesignSpecMetricBase` -- see comment above)?

src/xngin/apiserver/routers/common_api_types.py
```diff
@@ -137,6 +127,18 @@ class DesignSpecMetric(DesignSpecMetricBase):
...
+    @model_validator(mode="after")
+    def cluster_fields_check(self) -> Self:
+        """Enforce that cluster fields are either all set or all unset."""
+        cluster_fields = (self.icc, self.avg_cluster_size, self.cv)
+        if any(f is not None for f in cluster_fields) and any(f is None for f in cluster_fields):
+            raise ValueError("icc, avg_cluster_size, and cv must all be set together or all be None")
+        return self
+
+
+class DesignSpecMetric(DesignSpecMetricBase):
```
poornimaramesh: just to understand -- it looks like DesignSpecMetric doesn't do anything except add the stddev_check. Why not simply add this check to the `DesignSpecMetricBase` class?

poornimaramesh (changes requested):
Lookks good, but could you address the comments about the DesignSpecMetric above?

src/xngin/apiserver/routers/common_api_types.py
mdulieu: The stddev rule was duplicated. Done in <commit>: stddev_check now lives in DesignSpecMetricBase and the duplicate clause in check_baseline_stats is removed, so both subclasses share the one check.

I kept DesignSpecMetric as a distinct class even though it is now empty, for two reasons. First, it keeps request and response metrics as separate types: as siblings, a function annotated with DesignSpecMetric will not accept a DesignSpecMetricRequest, whereas if the response type were the base class itself, every request would satisfy it by inheritance. Second, the class name is the schema name in the OpenAPI spec, and the frontend client is generated from it, so removing the class would rename the schema and churn the generated code for identical runtime behavior.

Snehaaa18:
Hii! @mdulieu I just wanted to flag this PR #333 which I am working on to flatten the power check structure might conflict with your changes, but I am happy to discuss common grounds or suggestions to manage it

mdulieu:
> Hii! @mdulieu I just wanted to flag this PR #333 which I am working on to flatten the power check structure might conflict with your changes, but I am happy to discuss common grounds or suggestions to manage it

Hi @Snehaaa18, thanks for flagging! Yes, we touch the same files. My changes are mostly additive (optional baseline stats fields on the metric models, plus a skip branch in power_check), so whichever PR lands second should be a manageable rebase. My slight preference would be this one first, since a refactor rebases over additions more easily than the reverse, and #336 stacks on top of it, but I am fine either way. Happy to help resolve conflicts on either side.

src/xngin/apiserver/routers/admin/test_admin_api.py
```diff
@@ -3938,6 +3939,150 @@ def make_design_spec(*, desired_n: int | None = None, desired_n_clusters: int |
     assert both_analysis.pct_change_with_desired_n == clusters_analysis.pct_change_with_desired_n
 
 
+def echo_metric_request(spec: DesignSpecMetric) -> DesignSpecMetricRequest:
+    """Builds a follow-up power check metric from a prior response's metric_spec, as the frontend would."""
+    return DesignSpecMetricRequest(
+        field_name=spec.field_name,
+        metric_pct_change=spec.metric_pct_change,
```
qixotic: I'd add a comment that `metric_target` is explicitly not copied since it would be derived from the metric_pct_change.

src/xngin/apiserver/routers/common_api_types.py
```diff
@@ -137,19 +127,38 @@ class DesignSpecMetric(DesignSpecMetricBase):
...
+    """
+
+
 class DesignSpecMetricRequest(DesignSpecMetricBase):
-    """Defines a request to look up baseline stats for a metric to measure in an experiment."""
+    """Defines a request to look up baseline stats for a metric to measure in an experiment.
 
-    # TODO: consider supporting {metric_baseline, metric_stddev, available_n} as inputs when the metric may not exist or
-    # be usable yet in the dwh, so that it it can be used as a general power/sizing calculator.
+    Baseline stats may optionally be supplied (e.g. echoed back from a prior power check's
```
qixotic: I'd replace "echoed back" with "copied".

src/xngin/apiserver/routers/common_api_types.py
```diff
@@ -175,6 +184,39 @@ def check_has_only_one_of_pct_change_or_target(self) -> Self:
...
+            raise ValueError(
+                "metric_type, metric_baseline, available_nonnull_n, and available_n must all be set "
+                "together or all be None"
+            )
+        return self
+
+    @property
+    def has_baseline_stats(self) -> bool:
+        """True when this request carries the baseline stats needed to skip the dwh stats query."""
+        return self.metric_baseline is not None
```
qixotic: Shouldn't this check that all of (self.metric_type, self.metric_baseline, self.available_nonnull_n, self.available_n) are not None?

src/xngin/apiserver/routers/test_common_api_types.py
```diff
@@ -334,3 +336,70 @@ def call(label):
...
+        metric_pct_change=0.1,
+        metric_type=MetricType.NUMERIC,
+        metric_baseline=100.0,
+        metric_stddev=15.0,
+        available_nonnull_n=900,
+        available_n=1000,
+    )
+    assert full_stats.has_baseline_stats is True
+
+    # A partial set of baseline stats is rejected.
```
qixotic: I don't think this tests has_baseline_stats fully. The following behavior is what I should expect, yes?

```
    # A partial set of baseline stats is rejected.  First create without validation.
    partial_stats = DesignSpecMetricRequest.model_construct(
        field_name="metric1", metric_pct_change=0.1, metric_baseline=100.0
    )

    # If such a model were normally validated, it would raise a validation error:
    with pytest.raises(ValidationError, match="must all be set together"):
        DesignSpecMetricRequest.model_validate(partial_stats)

    # But if we used the unvalidated model, we can still use has_baseline_stats as a manual check.
    assert partial_stats.has_baseline_stats is False
```

src/xngin/apiserver/routers/admin/admin_api.py
```diff
@@ -1983,76 +1989,104 @@ async def power_check(
...
-            raw_cluster_stats = await asyncio.to_thread(
-                get_cluster_sufficient_stats,
-                dwh.session,
-                sa_table,
-                cluster_key,
-                db_derived_metrics,
-                filters,
-                outcome_shifts,
+    metrics_missing_stats = [m for m in design_spec.metrics if not m.has_baseline_stats]
+    needs_cluster_stats = cluster_key is not None and any(m.icc is None for m in design_spec.metrics)
```
qixotic: Shouldn't this check be more comprehensive, too, i.e. check that all of (self.icc, self.avg_cluster_size, self.cv) are not None? You could add a property similar to has_baseline_stats.

src/xngin/apiserver/routers/common_api_types.py
mdulieu: Done!

src/xngin/apiserver/routers/admin/test_admin_api.py
mdulieu: Done, added to the helper's docstring.

src/xngin/apiserver/routers/common_api_types.py
mdulieu: Yes, done! 
It now checks all four fields directly instead of relying on the validator invariant, so it also gives the right answer on objects built with model_construct.

src/xngin/apiserver/routers/admin/admin_api.py
mdulieu: Done: added a has_cluster_stats property that checks all three values, and updated the three places in power_check that previously only checked icc (the skip decision, the list of metrics to query, and the user-provided branch), so they all apply the same rule.

src/xngin/apiserver/routers/test_common_api_types.py
mdulieu: Added, thanks! One tweak: model_validate on a model instance skips re-validation by default in Pydantic v2, so now the test validates partial_stats.model_dump() instead, which does trigger the error.

qixotic (approved):

PR #328, opened 2026-09-02
Remove the concept of participant types.

dctaf:
## Ticket

EVE-199

## Description

Removes the concept of a "participant type" from the API, storage, and
implementation.

Some of the ideas from this code, such as drift detection and
`get_stats_on_filters`, will be restored later to support other features.

This PR will be followed by another cleanup PR to further simplify some types
and tighten some validation.


src/xngin/apiserver/dwh/queries.py
```diff
@@ -82,75 +75,6 @@ def get_stats_on_metrics(
     return metrics_to_return
 
 
-def get_stats_on_filters(
```
poornimaramesh, line 85: I'm wondering if we should re-use this function in `src/xngin/apiserver/dwh/dwh_session`'s `inspect_table`? It would be useful to check the filters before we apply them at experiment creation time...

poornimaramesh (approved):
Looks good to me, but might be worth @qixotic taking another look

src/xngin/apiserver/dwh/queries.py
dctaf: I agree. We can restore this later. There are opportunities to use this data to improve the interactive experiment filters UX.

markbotterill:
@qixotic do you want that second glance? I am just about to start tackling the other 2 PRs in the stack 🙏

qixotic (approved):
Sorry for the delay, LGTM, including the output of tools/migrations/eve-199-participant-type-scope.sql against prod! Thanks Doug and Poornima!

PR #332, opened 2026-09-07
fix outcome / context vals getting appended twice for calculations

poornimaramesh:
## Ticket

Fixes [EVE-171](https://idinsight.atlassian.net/browse/EVE-171)

## Description
Prevent logging outcomes and context values twice while updating the bandit arms

## To-do before merge
- Delete test in `test_experiments_api.py`

### DIFF
--- a/src/xngin/apiserver/routers/experiments/experiments_common.py
+++ b/src/xngin/apiserver/routers/experiments/experiments_common.py
@@ -1043,15 +1043,14 @@ async def _fetch_outcomes_and_context_for_arm(
     xngin_session: AsyncSession,
     experiment_id: str,
     arm_id: str,
-    outcome: float,
-    context_vals: list[float] | None,
+    experiment_type: ExperimentsType,
 ) -> tuple[list[float], list[list[float]] | None]:
     """Return the just-recorded outcome plus the arm's prior outcomes, newest first.
 
     Context values are aggregated alongside the outcomes and returned only for contextual
     bandits, i.e. when the current draw carries context.
     """
-    has_context = context_vals is not None
+    is_cmab = experiment_type == ExperimentsType.CMAB_ONLINE
     subq = (
         select(tables.Draw.outcome, tables.Draw.context_vals)
         .where(
@@ -1064,15 +1063,14 @@ async def _fetch_outcomes_and_context_for_arm(
         .subquery()
     )
     agg_cols = [func.array_agg(subq.c.outcome)]
-    if has_context:
+    if is_cmab:
         agg_cols.append(func.array_agg(subq.c.context_vals))
     agg_result = await xngin_session.execute(select(*agg_cols).select_from(subq))
     agg_row = agg_result.one()
 
-    all_prior_outcomes = agg_row[0]
-    outcomes = [outcome] + (all_prior_outcomes or [])
-    all_context_vals = ([context_vals] + (agg_row[1] or [])) if has_context else None
-    return outcomes, all_context_vals
+    all_outcomes = agg_row[0]
+    all_context_vals = agg_row[1] if is_cmab else None
+    return all_outcomes, all_context_vals
 
 
 class PartialUpdateDrawBeta(TypedDict):
@@ -1168,8 +1166,7 @@ async def update_bandit_arm_with_outcome_impl(
         xngin_session,
         experiment_id=experiment.id,
         arm_id=draw_record.arm_id,
-        outcome=outcome,
-        context_vals=draw_record.context_vals,
+        experiment_type=ExperimentsType(experiment.experiment_type),
     )
 
     updated_parameters = update_bandit_arm(
--- a/src/xngin/apiserver/routers/experiments/test_experiments_api.py
+++ b/src/xngin/apiserver/routers/experiments/test_experiments_api.py
@@ -1485,7 +1485,7 @@ async def test_update_bandit_arm_with_freq_experiments_returns_422(
     assert "Cannot dynamically update arms for frequentist experiments" in str(exc.value.result.data)
 
 
-@pytest.mark.skip("EVE-171")
+# TODO: delete this test before merging the PR
 async def test_normal_prior_binary_reward_fits_each_outcome_exactly_once(
     testing_datasource, aclient: AdminAPIClient, eclient: ExperimentsAPIClient
 ):


src/xngin/apiserver/routers/experiments/experiments_common.py
```diff
@@ -1043,15 +1043,14 @@ async def _fetch_outcomes_and_context_for_arm(
...
+    experiment_type: ExperimentsType,
 ) -> tuple[list[float], list[list[float]] | None]:
     """Return the just-recorded outcome plus the arm's prior outcomes, newest first.
 
     Context values are aggregated alongside the outcomes and returned only for contextual
     bandits, i.e. when the current draw carries context.
     """
-    has_context = context_vals is not None
+    is_cmab = experiment_type == ExperimentsType.CMAB_ONLINE
     subq = (
```
markbotterill, line 1054: Can I suggest we rename "subq" -> "last_100_draws" (or "last_n_draws" if we do implement the adjustable number of prior draws). I found this code quite hard to parse and a lot of that was simply because of the variable names!

src/xngin/apiserver/routers/experiments/experiments_common.py
```diff
@@ -1064,15 +1063,14 @@ async def _fetch_outcomes_and_context_for_arm(
         .subquery()
     )
     agg_cols = [func.array_agg(subq.c.outcome)]
```
markbotterill, line 1065: Similarly, can we change this to `outcomes_col = func.array_agg(recent_draws.c.outcome).label("outcomes")`

src/xngin/apiserver/routers/experiments/experiments_common.py
```diff
@@ -1064,15 +1063,14 @@ async def _fetch_outcomes_and_context_for_arm(
...
-    if has_context:
+    if is_cmab:
```
markbotterill, line 1066: And can we change this to 

```
if is_cmab:
          context_col = func.array_agg(recent_draws.c.context_vals).label("context_vals")
          row = (await xngin_session.execute(select(outcomes_col, context_col))).one()
          return row.outcomes, row.context_vals
```

src/xngin/apiserver/routers/experiments/experiments_common.py
```diff
@@ -1064,15 +1063,14 @@ async def _fetch_outcomes_and_context_for_arm(
...
     agg_result = await xngin_session.execute(select(*agg_cols).select_from(subq))
     agg_row = agg_result.one()
 
-    all_prior_outcomes = agg_row[0]
-    outcomes = [outcome] + (all_prior_outcomes or [])
-    all_context_vals = ([context_vals] + (agg_row[1] or [])) if has_context else None
-    return outcomes, all_context_vals
+    all_outcomes = agg_row[0]
+    all_context_vals = agg_row[1] if is_cmab else None
+    return all_outcomes, all_context_vals
```
markbotterill, line 1073: Full rewrite would look something along these lines:
```
async def _fetch_recent_outcomes_for_arm(
      xngin_session: AsyncSession,
      experiment_id: str,
      arm_id: str,
      experiment_type: ExperimentsType,
  ) -> tuple[list[float], list[list[float]] | None]:
      """Return the arm's recorded outcomes, most recently observed first, capped at 100.

      Index 0 is the outcome written earlier in this transaction. Context values are
      returned only for contextual bandits.
      """
      is_cmab = experiment_type == ExperimentsType.CMAB_ONLINE
      recent_draws = (
          select(tables.Draw.outcome, tables.Draw.context_vals)
          .where(
              tables.Draw.experiment_id == experiment_id,
              tables.Draw.arm_id == arm_id,
              tables.Draw.outcome.is_not(None),
          )
          .order_by(tables.Draw.observed_at.desc())
          .limit(100)  # TODO: Make draw limiting configurable
          .subquery()
      )
      outcomes_col = func.array_agg(recent_draws.c.outcome).label("outcomes")

      if is_cmab:
          context_col = func.array_agg(recent_draws.c.context_vals).label("context_vals")
          row = (await xngin_session.execute(select(outcomes_col, context_col))).one()
          return row.outcomes, row.context_vals

      row = (await xngin_session.execute(select(outcomes_col))).one()
      return row.outcomes, None
```

src/xngin/apiserver/routers/experiments/experiments_common.py
```diff
@@ -1043,15 +1043,14 @@ async def _fetch_outcomes_and_context_for_arm(
...
 ) -> tuple[list[float], list[list[float]] | None]:
     """Return the just-recorded outcome plus the arm's prior outcomes, newest first.
 
     Context values are aggregated alongside the outcomes and returned only for contextual
     bandits, i.e. when the current draw carries context.
     """
-    has_context = context_vals is not None
+    is_cmab = experiment_type == ExperimentsType.CMAB_ONLINE
     subq = (
         select(tables.Draw.outcome, tables.Draw.context_vals)
```
markbotterill, line 1055: I think there's a breaking change with this Query now: we order results by `.order_by(tables.Draw.created_at.desc())`. That USED to be fine because we used to prepend the latest result - so we always knew that the result[0] was gonna be the thing we were updating. Since we're dropping this logic, we now actually DO rank by the order of draw creation -> this doesn't necessarily map to the timing of observing an outcome! We could do a draw 1 at 10am, draw 2 at 10:30am -> draw 2 could report back at 10:31am with it's outcome, draw 1 could report back at 11am. So wouldn't that mean that we'd be misrepresenting draw 2's outcome as draw 2?

markbotterill: Sorry! Mistake in that:  Draw 2's outcome isn't misattributed to draw 2, it gets applied a second time, and draw 1's outcome never gets applied at all!

src/xngin/apiserver/routers/experiments/experiments_common.py
markbotterill: I just found "agg_cols" a bit confusing to read!

src/xngin/apiserver/routers/experiments/experiments_common.py
```diff
@@ -1168,8 +1166,7 @@ async def update_bandit_arm_with_outcome_impl(
         xngin_session,
```
markbotterill, line 1166: We would need to change the docstring just above as well - not sorting by creation date but by outcome reporting I think

src/xngin/apiserver/routers/experiments/test_experiments_api.py
```diff
@@ -1485,7 +1485,7 @@ async def test_update_bandit_arm_with_freq_experiments_returns_422(
     assert "Cannot dynamically update arms for frequentist experiments" in str(exc.value.result.data)
 
 
-@pytest.mark.skip("EVE-171")
+# TODO: delete this test before merging the PR
```
markbotterill, line 1488: Shouldn't we keep this? It's proof that we fix the prepend issue?

src/xngin/apiserver/routers/experiments/test_experiments_api.py
```diff
@@ -1485,7 +1485,7 @@ async def test_update_bandit_arm_with_freq_experiments_returns_422(
...
 async def test_normal_prior_binary_reward_fits_each_outcome_exactly_once(
     testing_datasource, aclient: AdminAPIClient, eclient: ExperimentsAPIClient
 ):
```
markbotterill, line 1491: Also should we add a test that checks for the older draw -> newer reported outcome issue I raised in above comments?

Claude suggesting an example here:
```
async def test_update_bandit_arm_applies_outcome_of_older_draw_reported_later(xngin_session, testing_datasource):
      """The outcome being recorded is the one applied, even when a newer draw already reported."""
      experiment = await insert_experiment_and_arms(
          xngin_session,
          testing_datasource.ds,
          experiment_type=ExperimentsType.MAB_ONLINE,
          prior_type=PriorTypes.BETA,
          reward_type=LikelihoodTypes.BERNOULLI,
      )
      older = await create_assignment_for_participant(xngin_session, experiment, "participant-1", random_state=66)
      newer = await create_assignment_for_participant(xngin_session, experiment, "participant-2", random_state=66)
      assert older is not None and newer is not None
      assert older.arm_id == newer.arm_id
      # Both draws share a created_at inside one transaction; push the first one into the past.
      await xngin_session.execute(
          update(tables.Draw)
          .where(tables.Draw.experiment_id == experiment.id, tables.Draw.participant_id == "participant-1")
          .values(created_at=datetime.now(UTC) - timedelta(hours=1))
      )
      arm = next(arm for arm in experiment.arms if arm.id == older.arm_id)
      assert arm.alpha is not None and arm.beta is not None
      initial_alpha, initial_beta = arm.alpha, arm.beta

      # The newer draw reports a success first, then the older draw reports a failure.
      await update_bandit_arm_with_outcome_impl(xngin_session, experiment, "participant-2", outcome=1.0)
      await update_bandit_arm_with_outcome_impl(xngin_session, experiment, "participant-1", outcome=0.0)

      assert arm.alpha == initial_alpha + 1
      assert arm.beta == initial_beta + 1

```

markbotterill: If you add this test in it should fail for the current code in this PR

markbotterill (changes requested):
This still leaves some issues for Normal-Normal + Beta MABs which I've flagged. Also suggested some variable name changes as I found it quite hard to pick this PR and code apart!

src/xngin/apiserver/routers/experiments/experiments_common.py
poornimaramesh: done, thanks

src/xngin/apiserver/routers/experiments/experiments_common.py
poornimaramesh: also fixed, thanks

src/xngin/apiserver/routers/experiments/experiments_common.py
poornimaramesh: fixed

src/xngin/apiserver/routers/experiments/experiments_common.py
poornimaramesh: done

src/xngin/apiserver/routers/experiments/experiments_common.py
poornimaramesh: done

src/xngin/apiserver/routers/experiments/test_experiments_api.py
poornimaramesh: added the test, but without the calls to the db

src/xngin/apiserver/routers/experiments/experiments_common.py
poornimaramesh: fixed, thanks for catching!

PR #333, opened 2026-09-07
Flatten Power_Request

Snehaaa18:
## Ticket

EVE-194 (https://idinsight.atlassian.net/browse/EVE-194)

## Description
Power checks are now handled directly from the power request rather than routed through a design spec. Callers only need to supply the limited set of fields power calculation actually uses, instead of constructing a full design spec.

## Checklist

Fill with `x` for completed.

- [ ] I have updated the automated tests


src/xngin/apiserver/routers/common_api_types.py
```diff
@@ -1232,7 +1232,37 @@ class CMABExperimentSpec(BaseBanditExperimentSpec):
...
+    n_arms: int
+    arm_weights: list[float] | None = None
+    power: float = 0.8
+    alpha: float = 0.05
+    desired_n: int | None = None
+    desired_n_clusters: int | None = None
+
+    @model_validator(mode="after")
+    def validate_arm_weights(self) -> Self:
+        """Replaces BaseDesignSpec.get_validated_arm_weights, which this flat model does not inherit.
```
qixotic, line 1248: I'd simplify this doc string to just note its core purpose similar to the original fn's description:

If arm_weights exist, validate that the length matches n_arms and weights sum to 100 before returning.

src/xngin/apiserver/routers/common_api_types.py
```diff
@@ -1232,7 +1232,37 @@ class CMABExperimentSpec(BaseBanditExperimentSpec):
 
 
 class PowerRequest(ApiBaseModel):
-    design_spec: AnyFrequentistDesignSpec
+    table_name: str
+    cluster_key: str | None = None
+    filters: list[Filter] = []
+    metrics: list[DesignSpecMetricRequest]
+    n_arms: int
+    arm_weights: list[float] | None = None
```
qixotic, line 1240: You dropped the old constraints from Arm.arm_weight. I think the equivalent would still leverage the exisint `ArmWeight` type like:

    arm_weights: Annotated[list[ArmWeight] | None,  Field(
            description=(...)] = None

src/xngin/apiserver/routers/common_api_types.py
```diff
@@ -1232,7 +1232,37 @@ class CMABExperimentSpec(BaseBanditExperimentSpec):
 
 
 class PowerRequest(ApiBaseModel):
-    design_spec: AnyFrequentistDesignSpec
+    table_name: str
+    cluster_key: str | None = None
+    filters: list[Filter] = []
+    metrics: list[DesignSpecMetricRequest]
+    n_arms: int
```
qixotic, line 1239: You dropped the min_length=2 constraint when flattening; I believe we want:

n_arms = Annotated[int, Field(gte=2, lte=MAX_NUMBER_OF_ARMS)]

src/xngin/apiserver/routers/common_api_types.py
```diff
@@ -1232,7 +1232,37 @@ class CMABExperimentSpec(BaseBanditExperimentSpec):
 
 
 class PowerRequest(ApiBaseModel):
-    design_spec: AnyFrequentistDesignSpec
+    table_name: str
```
qixotic, line 1235: Please reuse/adapt any description and constraints applied from the original fields we've extracted here and below to this power request use case, as these get surfaced in the generated docs for developers looking to understand how to code to the API, thanks! (I've given a few examples below but not for all the fields.)

src/xngin/apiserver/routers/experiments/experiments_common.py
qixotic, line 202: Since power requests reuse this helper but no longer use a design_spec, this error message needs updating. How about making it a little more general, like:
"Your request refers to columns..."

src/xngin/apiserver/routers/experiments/experiments_common.py
```diff
@@ -237,6 +237,39 @@ def convert_table_to_fields_or_raise(table: Table, design_spec: AnyFrequentistDe
...
+    cluster_key: str | None = None,
+) -> dict[str, DataType]:
+    referenced_fields = {
+        *[metric.field_name for metric in metrics],
+        *[filter_.field_name for filter_ in filters],
+    }
+    if cluster_key is not None:
+        referenced_fields.add(cluster_key)
+
+    referenced_fields_and_types = _resolve_referenced_field_types(table, referenced_fields)
```
qixotic, line 254: From this line onward you appear to reuse the same code as the original convert_table_to_fields_or_raise. To avoid drift in the checks, let's extract another helper function that both ths old and this new fn can use. I think you'd need as params:
* referenced_fields_and_types
* metrics
* filters

qixotic (changes requested):
API docs and constraints need a little more cleanup. Please also reference the coupled FE change that needs to merge with this BE change in https://github.com/agency-fund/evidential-fe/pull/311. Thanks!

qixotic (approved):
Thanks!

PR #334, opened 2026-09-08
Compute cluster ICC/CV from per-cluster sufficient statistics in SQL

mdulieu:
Compute cluster ICC/CV from per-cluster sufficient statistics in SQL instead of shipping every row to Python.

**Stacked on #321** — the first commit here belongs to that PR; only the last
commit is under review. The diff collapses to just this change once #321 merges.

## What changed

- `get_cluster_outcome_data` is replaced by `get_cluster_sufficient_stats`: one
  `GROUP BY cluster` query returning `count(*)` per cluster plus
  `count/sum/sumsq` per metric. One row per cluster crosses the wire instead of
  one per participant.
- New pure function `calculate_icc_from_sufficient_stats` reconstructs the same
  one-way ANOVA ICC exactly from those triples. The query and the math are now
  fully decoupled: only the fetch touches the DWH session.
- pandas is no longer needed in `power_adapters.py`.
- Null semantics are unchanged: cluster sizes/CV count all rows (`count(*)`),
  ICC uses each metric's non-null values (`count(x)`/`sum(x)` skip nulls).

## Numerical stability

`sum(x*x)` loses precision when a metric's mean dwarfs its variance. Each
metric's mean (already fetched by `get_stats_on_metrics`) is passed as a shift
subtracted in SQL before summing; ICC is shift-invariant, so results are
unchanged (covered by tests, including a mean-1e9 case).

## Tests

- Equivalence tests: the sufficient-stats estimator matches
  `calculate_icc_from_dataframe` on unequal clusters, binary outcomes, and
  shifted data.
- All pre-existing `test_power_adapters.py` tests pass unchanged against
  Postgres, including the exact-value null-handling test.

### DIFF
--- a/src/xngin/apiserver/dwh/queries.py
+++ b/src/xngin/apiserver/dwh/queries.py
@@ -1,4 +1,4 @@
-from collections.abc import Sequence
+from collections.abc import Mapping, Sequence
 
 import sqlalchemy
 from sqlalchemy import (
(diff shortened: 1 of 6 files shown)


poornimaramesh (approved):

PR #335, opened 2026-09-08
Separate DWH queries from stats computation in power_check

mdulieu:
Implements the query/compute separation suggested in the #321 review:
code inside a `DwhSession` block only queries; computation on the results
happens after the block (and the remote connection) is closed.

**Stacked on #334** — the first two commits belong to #321/#334; only the last
commit is under review. The diff collapses to just this change once they merge.

## What changed

- `queries.py` now contains only queries. `get_stats_on_metrics` became
  `get_raw_metric_stats`, returning the raw aggregate row.
- `power_adapters.py` is now entirely session-free: `build_metric_stats` turns
  the raw row into `DesignSpecMetric` objects, and `calculate_cluster_stats`
  computes ICC/CV from the per-cluster sufficient-statistics rows.
- `power_check` fetches inside the `DwhSession` block and computes after it,
  matching the pattern from the review.

No behavior change; all existing tests pass unchanged (test files compose
query + compute through small helpers with the old signatures).


poornimaramesh (approved):

PR #336, opened 2026-09-09
Return an MDE curve from power checks for a requested list of sample sizes

mdulieu:
Power checks can now return a full MDE-vs-sample-size curve in a single request. The design
spec accepts an optional list of desired sample sizes (or cluster counts), and the response
includes the minimum detectable effect for each one, so the frontend can plot how detectable
effect improves as the sample grows — without issuing one request per point or re-querying
the data warehouse.

**Stacked on #327** — the first two commits belong to that PR; only the last commit is under
review here. The diff collapses once #327 merges.

## API change

- `desired_ns` (frequentist design specs) and `desired_ns_clusters` (preassigned specs, takes
  precedence, requires `cluster_key`): optional lists of desired sample sizes / cluster counts,
  each entry ≥ 1, at most 100 entries.
- `MetricPowerAnalysis.mde_curve`: one `MdeCurvePoint {desired_n, desired_n_clusters,
  pct_change}` per requested size, in request order. Cluster counts convert per metric via
  `avg_cluster_size`, exactly like the single `desired_n_clusters` (#322).

## Behavior

- Each point is computed best-effort: a size the solver can't handle (e.g. too small) yields a
  null `pct_change`; the request never fails on a single point.
- The point list is caller-defined, so point count and spacing are entirely FE choices — no API
  change needed to tune the chart later.
- Combined with #327's stats reuse, a follow-up curve request with echoed stats computes all
  points with **zero** dwh queries (covered by a test that proves it with a nonexistent table
  name).

## Tests

Stats-level (ordering, per-metric cluster conversion, null-on-failure, consistency with the
single-value `pct_change_with_desired_n`), model validation (`desired_ns_clusters` requires
`cluster_key`), and endpoint-level for both individual and cluster designs.

## Follow-up (FE)

Chart in the experiment wizard: 10 points spaced uniformly in MDE (√-spacing) from the minimum
sample size across `available_n`, threshold lines at min and available n.

### DIFF
--- a/src/xngin/apiserver/limits.py
+++ b/src/xngin/apiserver/limits.py
@@ -40,6 +40,9 @@ MAX_NUMBER_OF_FIELDS = 150
 # Maximum number of filters we allow to define an audience.
 MAX_NUMBER_OF_FILTERS = 20
 
+# Maximum number of desired sample sizes a power check may request MDEs for in one call.
+MAX_NUMBER_OF_POWER_CURVE_POINTS = 100
+
 # Maximum length of GCP service account info JSON.
 MAX_GCP_SERVICE_ACCOUNT_LEN = 8000
 
--- a/src/xngin/apiserver/routers/admin/admin_api.py
+++ b/src/xngin/apiserver/routers/admin/admin_api.py
@@ -2248,7 +2248,13 @@ async def power_check(
     user: Annotated[tables.User, Depends(require_user_from_token)],
     body: PowerRequest,
 ) -> PowerResponse:
-    """Performs a power check for the specified datasource."""
+    """Performs a power check for the specified datasource.
+
+    Metrics carrying baseline stats (e.g. echoed back from a prior power check response) are used
+    as-is rather than re-queried from the data warehouse. When no metric needs the warehouse — every
+    metric has baseline stats, and cluster stats for cluster-randomized designs — the warehouse is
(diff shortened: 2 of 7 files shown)


qixotic:
I merged in main and added a few more tiny tests; LGTM.

Also as discussed, we're primarily keeping the calculation in the backend instead of scaling a reference MDE by sqrt(n_ref / n) (without implementing an equivalent solver in the FE) because:
* Numeric metrics use a t-test, whose critical value changes with n. Although the t-test converges to the z-test as n grows, the effect size is not really proportional to 1/sqrt(n) as n gets small and the t critical value drifts more from the z critical value, resulting in a plot that sits below the actual curve generated by the BE. And clustered designs due to having an effective size = n / design effect would hit this behavior even sooner.
* Although binary metrics _could_ reproduce a curve from a reference MDE, scaling a raw percent would be incorrect. The FE would have to convert to Cohen's h, scale h by sqrt(n_ref / n) using control arm counts for both n's, and convert back to in the same direction as the BE (downward today, but note https://github.com/agency-fund/evidential-be/pull/345). A point where φ(baseline) +/- h goes beyond [0, π] has to be dropped, because even a swing to 0% or 100% in that direction misses the power target (80% by default). Keeping the curve on the backend uses the same solver that produces pct_change_with_desired_n at the user's selected sample size, so that value is exactly the curve point there.

qixotic (approved):

PR #337, opened 2026-09-09
Revamp how REST resources are loaded from the database.

dctaf:
# Ticket

EVE-206

# Description

Refactors resource loading in our APIs:

- Moves authentication, authorization, resource lookup, and missing-resource handling into named dependencies.
- Adds shared helpers and constants for direct and nested SQLAlchemy preloads.
- Adds a test comparing route path parameters with those consumed by handlers and transitive dependencies.
- Enforces `adeps`, `ideps`, and `edeps` import conventions with Ruff.
- Preserves existing API behavior.


markbotterill:
> ⚠️ This PR requires frontend changes: client generation or linting failed. Please review: https://github.com/agency-fund/evidential-be/actions/runs/34672783827

Is this anything to be concerned by? Seeing a lot of issues around `ParticipantTypes` in the lint run.

src/xngin/apiserver/test_handler_routes.py
```diff
@@ -0,0 +1,92 @@
+"""Checks that every route's path parameters line up with what its handler and dependencies consume.
```
markbotterill, line 1: Took me a while to parse through these 😮‍💨 but really glad we have this test in!
 
I am a bit confused on one thing; the second bullet of the docstring says the check catches a handler that "drops the organization constraint." It catches a handler that drops the `organization_id` name. From what I understand, the check here is name-level, same as FAST003. This is totally fine and makes sense, but I was noodling about whether we should be worried about going deeper. 

Whether the dependency that consumed a child id also constrained it to its parent (datasource to org, experiment to datasource, snapshot to experiment) is a fact about the query inside each dependency, which neither check can see (because they're name only). I imagine the test required to actually check this in general might be a bit more involved, but as things stand it seems like we could have two dependencies that check different things with no check that they are actually interlinked. 

Org + datasource is the case closest to hand (and we have  `_org_datasource` for exactly this reason!), since it is the one resource with two loaders: a handler that takes `adeps.organization` and `adeps.experiment` side by side consumes every name and passes, but never confirms the datasource belongs to the org in the URL.

I suppose I am thinking about protecting for a more generic future case case - like, if we later add per-org admins, a delete route written that way lets an admin of org A delete a snapshot in org B's datasource by putting A in the URL, as long as they are a regular member of B.

Is it worth a sentence in the docstring saying the check is name-level and the parent-child constraints will rely on coding out / picking the right chained dependency (e.g. `_org_datasource`)? Or is this all getting ahead of things a bit / am I barking up wrong tree?

src/xngin/apiserver/routers/admin/admin_dependencies.py
markbotterill, line 1: Small note on the file as a whole, would you be open to re-structuring it a bit? Just found I had to hop up and down while reviewing. The following would still satisfy Python constraints of making sure things are defined in correct order while also being slightly easier to scan (at least to me!).
```
  # Organization
  load_organization_or_raise
  _Organization + instances

  # Datasource
  _load_datasource_or_raise
  _Datasource + instances
  _org_datasource

  # Experiment
  _load_experiment_or_raise
  _Experiment + instances
  org_experiment
```

markbotterill: Otherwise all the logic looks great!

src/xngin/apiserver/routers/integrations/integrations_api.py
```diff
@@ -102,7 +79,7 @@ async def get_turn_app_config(
     },
 )
 async def receive_turn_journey_update_notification(
-    webhook_id: str,
+    turn_webhook: Annotated[tables.Webhook, Depends(ideps.turn_webhook)],
     session: Annotated[AsyncSession, Depends(xngin_db_session)],
     auth_token: Annotated[str | None, Header(alias=constants.HEADER_WEBHOOK_TOKEN)] = None,
```
markbotterill, line 84: Do we still need the `auth_token` here? ` ideps.turn_webhook ` already reads this header.

Could we use the turn_webhook.auth_token  at line 102 now instead, i.e. `webhook_auth_token=turn_webhook.auth_token`?

src/xngin/apiserver/routers/admin/admin_dependencies.py
```diff
@@ -0,0 +1,337 @@
...
+) -> tables.Organization:
+    """Reads the requested organization from the database. Raises 404 if disallowed or not found.
+
+    Privileged users may access any organization; non-privileged users must be a member.
+    """
+    stmt = select(tables.Organization).where(tables.Organization.id == organization_id)
+    if not user.is_privileged:
+        stmt = stmt.join(tables.UserOrganization).where(tables.UserOrganization.user_id == user.id)
+    if preload:
+        stmt = stmt.options(*[selectinload(f) for f in preload])
```
markbotterill, line 84: Isn't this a place we could use `build_preload_options` too? I see we use it below at L148:`options = build_preload_options(preload, nested_preload)`

src/xngin/apiserver/routers/admin/admin_dependencies.py
```diff
@@ -0,0 +1,337 @@
...
+        .join(tables.Organization)
+        .join(tables.UserOrganization)
+        .where(
+            tables.UserOrganization.user_id == user.id,
+            tables.Datasource.id == datasource_id,
+        )
+    )
+    if organization_id:
+        stmt = stmt.where(tables.Organization.id == organization_id)
+    if preload:
```
markbotterill, line 118: Same preload Q here

markbotterill (approved):
This is great @dctaf - it tidies up so much code and makes things a lot neater! I have left some minor comments and clarifications but I am happy giving this a provisional approval as I don't think any are blocking. 

The only residual nit I have is around the naming of the `deps`; is it convenention to only use 1-2 letters for a prefix to `-deps` (i.e. `adeps`). To me, on first scan, I was wondering if the "a" was something to do with `async` or the like. I know they appear in a lot of places (so this adds a few characters) but what are thoughts on renaming to `adm_deps`, `exp_dep`, `int_deps`? To me it's a little more legibile.

UPDATE: While reviewing the next PR I can see that we have all the "aclient" and "eclient" etc. set up across the codebase, which gives me pause for thought. I stand by the original reaction of it feeling a little unintuitive as a naming convention, but addressing it would be quite a far-reaching change!

src/xngin/apiserver/routers/integrations/integrations_api.py
dctaf: Good catch. Fixed.

src/xngin/apiserver/routers/admin/admin_dependencies.py
dctaf: Done!

src/xngin/apiserver/test_handler_routes.py
dctaf: This is a great line of questioning! Yes, the implementation is only matching names. I've reworded this entire section to simplify the explanation and added a note that this doesn't absolve the developer from thinking of the semantics of hierarchical collections.

We could move more towards the generic future case eventually. Right now, we don't have canonical REST names (paths) for all resources for historical and usability reasons (example: the Experiments API refers to experiments under `/experiments/X`, but Admin API uses a hierarchical name).

src/xngin/apiserver/routers/admin/admin_dependencies.py
dctaf: Done!

src/xngin/apiserver/routers/admin/admin_dependencies.py
dctaf: Done!

dctaf:
> > ⚠️ This PR requires frontend changes: client generation or linting failed. Please review: https://github.com/agency-fund/evidential-be/actions/runs/34672783827
> 
> Is this anything to be concerned by? Seeing a lot of issues around `ParticipantTypes` in the lint run.

That failed because the evidential-fe main branch still had the participant type editor UI, which was removed in https://github.com/agency-fund/evidential-fe/pull/306. The linting now passes: [sticky comment](https://github.com/agency-fund/evidential-be/pull/337#issuecomment-5608160214), [workflow run](https://github.com/agency-fund/evidential-be/actions/runs/35492452153/job/106029639629?pr=337).

PR #340, opened 2026-09-12
Remove asyncio and add basic DWH timeouts.

dctaf:
## Ticket

EVE-168, EVE-205

## Description

Adds timeouts to DWH interactions and removes asyncio from all SQLAlchemy
interactions.

- Removes async from application database sessions, FastAPI handlers and dependencies,
  CSV paths, most background jobs, and tests.
- Runs each customer-DWH session on one helper thread under a shared deadline
  and maps DWH timeouts to HTTP 504 responses.
- Removes the superseded async DWH/session code and SQLAlchemy `asyncio` dependency.
- Cleans up naming.


src/xngin/apiserver/database.py
```diff
@@ -59,37 +58,35 @@ def get_sqlalchemy_database_url():
...
-    async_engine = create_async_engine(
+    engine = create_engine(
         database_url,
         connect_args={"application_name": APP_DB_APPLICATION_NAME},
-        execution_options={"logging_token": "app_async"},
+        execution_options={"logging_token": "app"},
         logging_name=SA_LOGGER_NAME_FOR_APP,
     )
 
     # We use expire_on_commit for reasons described in docs/SQLALCHEMY.md.
```
markbotterill: Went poking around the docs here to get some more info; I think we'll need to update it since the file mentioned here talks about "Our base table definition class includes
AsyncAttrs
so that you may read attributes using async I/O as needed.
"

src/xngin/apiserver/database.py
```diff
@@ -59,37 +58,35 @@ def get_sqlalchemy_database_url():
...
-@contextlib.asynccontextmanager
-async def setup():
+@contextlib.contextmanager
+def setup():
     global _GLOBAL_STATE
 
     database_url = get_server_database_url()
 
-    async_engine = create_async_engine(
+    engine = create_engine(
```
markbotterill: By using the default here with no pooling configs, the [docs](https://docs.sqlalchemy.org/en/21/core/engines.html#pooling) say we max out at 5 workers (with 10 extra) so 15 max.
Should we: 
1. Be eplicit about this
2. Raise it?

src/xngin/apiserver/flags.py
```diff
@@ -33,11 +33,35 @@ def truthy_env(env_var: str):
...
+def _get_nproc() -> int:
+    """Concurrent threads to use in batch jobs.
+
+    Defaults to 4. Override by setting the XNGIN_WORKERS environment variable.
+    """
+    workers = os.getenv(ENV_WORKERS, "").strip()
+    if workers:
+        return int(workers)
+    cpu_count = os.process_cpu_count() or 1
+    return max(4, cpu_count - 1)
```
markbotterill: On a 32-thread processor, with the max pool size in the DB set at 15, we would have anything after the 15th worker just timing out right?

src/xngin/ops/threads.py
```diff
@@ -0,0 +1,103 @@
...
+        if self.expired:
+            raise TimeoutError("The deadline passed before this call could start.")
+        future = self.submit(fn, *args, **kwargs)
+        try:
+            return future.result(timeout=remaining)
+        except TimeoutError:
+            if future.done():
+                # fn raised TimeoutError itself; the deadline is still intact.
+                raise
+            self._expired = True
```
markbotterill: I think, `_expired` never has any use. If it ever is set, it is set _after_ the deadline has passed, which means that line `return self._expired or self.remaining <= 0` would already returne true based on the `self.remaining`

PR #342, opened 2026-09-15
Clamp NaN predicted-mean CIs to zero-width for arms with all-identical outcomes

mdulieu:
## Problem

test_freq_experiments_analyze failed intermittently (a few percent of runs). When random arm assignment placed all-identical outcomes into one arm (e.g. all zeros of a binary metric), the predicted-mean variance for that arm is mathematically zero, but statsmodels computes it as Var(Intercept) + Var(coef) + 2*Cov, and floating-point cancellation can leave it slightly negative (about -1e-17). sqrt() of that yields NaN mean_ci_lower/mean_ci_upper, which serialize to null in the API and broke the test's not-None assertions. Real users with an all-identical arm hit the same null CIs.

## Fix

In analyze_experiment, after summary_frame(), replace NaN mean CI bounds with the predicted mean, giving a zero-width CI. This is the statistically correct result for an arm with zero outcome variance. An arm with no data at all keeps NaN bounds, since its mean is itself NaN, so the "no data" case is not masked.

## Testing

- New regression test test_analysis_with_all_identical_outcomes_in_one_arm uses a 7-participant dataset (baseline arm [1, 0, 0], treatment arm all zeros) verified to deterministically produce the negative variance. It fails without the fix and passes with it.
- Full stats suite passes.
- The previously flaky test_freq_experiments_analyze passed 40 out of 40 consecutive runs with the fix.


src/xngin/stats/analysis.py
```diff
@@ -128,6 +128,12 @@ def analyze_experiment(
         pred_input = pd.DataFrame({arm_col: arm_ids})
         predictions = model.get_prediction(pred_input)
         pred_summary = predictions.summary_frame(alpha=alpha)
+        # When all outcomes in an arm are identical, the predicted-mean variance is mathematically
+        # zero, but floating-point cancellation can make statsmodels compute it as slightly
+        # negative, so sqrt() turns the CI bounds into NaN. The true CI is zero-width, so collapse
+        # NaN bounds to the predicted mean (a NaN mean itself stays NaN).
+        for ci_col in ("mean_ci_lower", "mean_ci_upper"):
+            pred_summary[ci_col] = pred_summary[ci_col].fillna(pred_summary["mean"])
```
poornimaramesh: this code seems to assume that all NaN values for CI bounds are due to the identical outcomes problem -- is that correct? Shouldn't we explicitly check for identical data for the arm, and _then_ backfill NaN values? -- otherwise this is an error that should be surfaced to the user no?

poornimaramesh (approved):
Approving, since the fix works, thanks! If you could address the comment above before merging though, that would be great!

src/xngin/stats/analysis.py
mdulieu: @poornimaramesh An arm with no data at all was already safe: its mean is NaN, so the backfill left the bounds NaN and they still surfaced as null. But you are right that a NaN variance with a finite mean can indeed arise from causes other than identical outcomes, for example zero residual degrees of freedom, and the original code would have masked those as a zero width CI. Updated: the backfill now applies only when the arm's outcomes are verifiably all identical with at least two observations (a single observation is trivially identical but carries no variance information). All other NaNs pass through to the caller as null. Added a regression test where one participant per arm makes the covariance inestimable and asserts the NaNs survive.

Could you review the changes before I merge? 
Thank you

